PR #6, opened 2015-07-02
Method signature changed and callback params for improve functionality and extensibility for i18n and more smart feedback component

vorjdux:
Hi Marc,

I changed a bit the component operation to make it more extensible and better prepared for an increase of more parameters and i18n.

Changed the input method and the first callback parameter to improve the ability to have more options in the future and let developers customize the titles of buttons.

Best regards.


vorjdux:
Hi, you saw the new code? Do you have some complaint or comment?

PR #10, opened 2015-07-17
Adding ability to return base64 image with image orientation.

carlykube:


marcshilling:
Cool, I didn't know you could get the base64 for library images too! Thanks!

Issue #18, opened 2015-08-19
Cannot read property 'showImagePicker' of undefined


klausbreyer:
Not able to implement the image picker. It seems like UIImagePickerManager is not defined.. 

console.log(require('NativeModules'));
=> Object {..}
ActionSheetManager: Object
ActivityIndicatorViewManager: Object
AlertManager: Object
AppState: Object
AsyncLocalStorage: Object
CameraManager: Object
CameraRollManager: Object
ContextExecutor: Object
DatePickerManager: Object
DevMenu: Object
EventDispatcher: Object
ExceptionsManager: Object
HTTPRequestHandler: Object
ImageLoader: Object
ImagePickerIOS: Object
ImageRequestHandler: Object
ImageStoreManager: Object
ImageViewManager: Object
LinkingManager: Object
LocationObserver: Object
MapManager: Object
NavItemManager: Object
NavigatorManager: Object
Networking: Object
PerfStats: Object
PerformanceLogger: Object
PickerManager: Object
ProgressViewManager: Object
RawTextManager: Object
Reachability: Object
ScrollViewManager: Object
SegmentedControlManager: Object
SettingsManager: Object
SliderManager: Object
SourceCode: Object
StatusBarManager: Object
SwitchManager: Object
TabBarItemManager: Object
TabBarManager: Object
TextFieldManager: Object
TextManager: Object
TextViewManager: Object
Timing: Object
UIManager: Object
Vibration: Object
ViewManager: Object
WebSocketExecutor: Object
WebSocketManager: Object
WebViewExecutor: Object
WebViewManager: Object
__proto__: Object

console.log(UIImagePickerManager);
=> undefined 

Here is my code: https://gist.github.com/klausbreyer/bdafe5ab0638bc60cbeb
And Xcode integration see screenshots below: 
![bildschirmfoto 2015-08-19 um 20 58 34](https://cloud.githubusercontent.com/assets/32771/9366207/3c8ae376-46b6-11e5-81d9-83907aa0fa93.png)
![bildschirmfoto 2015-08-19 um 20 52 02](https://cloud.githubusercontent.com/assets/32771/9366208/3c8e9df4-46b6-11e5-8394-fbd6aeb23da6.png)

error: 

![bildschirmfoto 2015-08-19 um 20 58 43](https://cloud.githubusercontent.com/assets/32771/9366206/3c8a87a0-46b6-11e5-8acc-5225d5dab304.png)

klausbreyer:
[...]

PR #19, opened 2015-08-20
cleared things up on the installation steps

klausbreyer:
I explicitly stated what Step I missed when I created #18


marcshilling:
So, this step isn't always necessary. I'm not sure what happened when you added the library, but I just made a fresh project, npm installed react-native-image-picker, added the folder to my project, and UIImagePickerManager.m was automatically added under 'Compile Sources' for me.

I'll update the README to include this as an optional step.

PR #25, opened 2015-09-16
Modal fix and max dimensions

udfalkso:
I've reworked this a bit, if you like the changes please accept the PR.  I don't think it negatively affects any current functionality.  

1)  Fixed it so that it works inside a RN Modal as well as outside one.
2)  Added support for maxWidth and maxHeight options that will downscale the image if needed before returning.
3)  Save the selected image to tmp directory and return a file:// style image url rather than an assets-library url.  This allows you to do things like upload it from objective c later with less trouble.  Also necessary for the downscaling.  

Apologies for the formatting changes, automatically done by my editor...

### DIFF
--- a/UIImagePickerManager/UIImagePickerManager.m
+++ b/UIImagePickerManager/UIImagePickerManager.m
@@ -1,5 +1,4 @@
 #import "UIImagePickerManager.h"
-#import <AssetsLibrary/AssetsLibrary.h>
 #import "RCTConvert.h"
 
 @interface UIImagePickerManager ()
--- a/UIImagePickerManager/UIImagePickerManager.m
+++ b/UIImagePickerManager/UIImagePickerManager.m
@@ -1,4 +1,5 @@
 #import "UIImagePickerManager.h"
+#import <AssetsLibrary/AssetsLibrary.h>
 #import "RCTConvert.h"
 
 @interface UIImagePickerManager ()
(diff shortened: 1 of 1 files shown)


udfalkso:
Oh, one thing I forgot to note.  This version has allowsEditing = true hardcoded.  Some people may not want this option.. should probably be made configurable at some point.

marcshilling:
I noticed that...I'll make it configurable. Thanks again!

udfalkso:
Awesome, thanks for merging in!

marcshilling:
@udfalkso so I merged this without fully testing...I can't get the image to display when using the `file:///` response. Does something special have to be done with that? I'm passing source to the `<Image>` as `{uri: file:///Users/marcshilling/Library/Developer/CoreSimulator/Devices/B66423FD…1-E90F-474A-BBDA-F2B0C29C7F6D/tmp/4EEDBD45-C52E-43C1-BAE9-CEE6D893FACD.jpg}`

udfalkso:
Ah sorry, I should have mentioned.  This seems like a RN bug... Here's what I do in my Image wrapper:

  if (source.uri.indexOf('file:///') === 0) {
        source.isStatic = true;
        source.uri = source.uri.replace('file://', '');
}

marcshilling:
Ahhh, got it. Cool, thanks again man.

PR #28, opened 2015-09-20
Add option to save image to documents dir

levic92:
Add storageOptions to allow saving to a path in the documents directory instead of tmp (which can get cleared).

```js
var options = {
  title: 'Select Photo',
  cancelButtonTitle: 'Cancel',
  takePhotoButtonTitle: 'Take Photo...',
  takePhotoButtonHidden: false,
  ...
  storageOptions: { // if this is provided image will get saved in the documents directory
    skipBackup: true, // will set attribute so the image is not backed up by icloud
    path: "images", // save in a sub directory in documents instead of root
  },
  ...
```


marcshilling:
Nice addition, thanks! This options object is starting to get out of hand but I guess mostly everything is optional so it doesn't matter haha.

PR #29, opened 2015-09-21
Fix presentation on iPad

levic92:
While testing on the simulator this was not working. It looks like the iPad uses an alert view instead of action sheet. Changed from using clickedButtonAtIndex to didDismissWithButtonIndex. This fixed the problem.

Tests performed:
Not in modal on iPhone: pass
In modal on iPhone: pass
Not in modal on iPad: pass
In modal on iPad: FAILED (also failed before this change)

Also found this article: http://stackoverflow.com/questions/25510894/prevent-my-ios-8-ipad-actionsheet-from-dismissing-its-parent-viewcontroller
"Moving forward you should embrace UIAlertController."


marcshilling:
Good catch. I'm going to merge this then look into changing it to `UIAlertController`.

PR #38, opened 2015-10-18
Added support for action sheets without titles

hyl:
Hi! 
Just a small PR to add support for having an action sheet without a title. At the moment I've added it simply as an option (```showTitle```): may be an idea to interpret  ```''``` or ```null``` as "don't display a title field" as well. 

showtitle=false
*<b>Above:</b> ```title=''```*

showtitle=true
*<b>Above:</b> ```showTitle=false```*

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ var UIImagePickerManager = require('NativeModules').UIImagePickerManager;
 
   // Specify any or all of these keys
   var options = {
+    showTitle: true,
     title: 'Select Avatar',
     cancelButtonTitle: 'Cancel',
     takePhotoButtonTitle: 'Take Photo...',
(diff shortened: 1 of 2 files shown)


marcshilling:
I'd prefer to just check for `''` or `null` like you said rather than adding a new option. If you rework it like that I'll merge.

marcshilling:
Actually, I'm about to open the code to change something anyway, so I'll do this right now.

marcshilling:
This is live in version `0.8.0` on npm. Thanks for the idea! I actually used this idea to remove some other redundant options keys. For example, rather than having a `takePhotoButtonHidden` option, just specify `takePhotoButtonTitle` as null or an empty string.

hyl:
Fantastic! Thanks very much. Happy to help! :) :+1:

PR #42, opened 2015-10-23
Thumbnail support

nevir:
Adds support for generating thumbnails in addition to the primary image:

```js
{
  maxWidth: 1024,
  maxHeight: 1024,
  thumbnail: {
    maxWidth: 128,
    maxHeight: 128,
    cover: true
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ var UIImagePickerManager = require('NativeModules').UIImagePickerManager;
       cover: true, // Whether the thumbnail should be cropped to exactly fit maxWidth & maxHeight.
     },
     allowsEditing: false, // Built in iOS functionality to resize/reposition the image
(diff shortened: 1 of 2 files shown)


marcshilling:
This is a cool PR, but it's kind of overkill for my goals with this library. User could always just specify maxWidth/maxHeight normally if they ONLY want a thumbnail. If they want both, there's nothing stopping the user from generating the thumbnail client side after the fact anyway.

nevir:
My main concern with generating thumbnails client side is that the base64 + string serialization is incredibly slow (a 1024x1024 image takes a good 200ms just to serialize to JS in the simulator on my MBP; let alone an actual device). 

Doing the image processing in js is equally cumbersome

Issue #48, opened 2015-11-04
Android support


whokilledtheelectricmonk:
It would be great to have support for this library on Android, using a native module or the [React Native Camera Roll API](https://facebook.github.io/react-native/docs/cameraroll.html)

marcshilling:
Currently no plans to work on this, but would gladly accept a PR!

nikki93:
Hey! I created a very basic (no extra options supported yet and has only `.launchCamera(...)` or `.launchImageLibrary(...)` directly) Android implementation of this API. You can find it here: https://github.com/exponentjs/react-native-image-picker-android

marcshilling:
@nikki93 sweet! Eventually we should try to merge these 2

marcshilling:
Merged in #54! Thanks @nikki93 and @yfuks!

PR #54, opened 2015-12-01
Android support

yfuks:
## Add android support based on [nikki93 works](https://github.com/exponentjs/react-native-image-picker-android)

***Issue*** : #48

### Information
- ImagePicker is render with a native android [dialog](http://developer.android.com/guide/topics/ui/dialogs.html)
- support **launchCamera()**, **launchImageLibrary()** and **showImagePicker()**

### values returned in case of success
- **path** : 'raw' path
- **uri** : real path
- **data** : the base64 encoded image data

### limitations
- only title option is supported
- only returns path, uri and data
- only support _android 3.0_ and higher

### postponed
- options cancelButtonTitle, takePhotoButtonTitle and chooseFromLibraryButtonTitle
- options noData
- options maxWidth and maxHeight


marcshilling:
Awesome stuff!

yfuks:
thanks !

nikki93:
@yfuks I think in the line https://github.com/marcshilling/react-native-image-picker/blob/master/Android/src/main/java/com/imagepicker/ImagePickerModule.java#L248 you need to use the URI and not just the file path directly, so you will need the equivalent of `response.putString("uri", Uri.fromFile(new File(realPath)).toString());`.

yfuks:
To tell you the truth i never code in JAVA before this module, is just try to make it as uselfull as i can, and the Uri.toString() basically return something like this:
```java
content://media/external/video/media
```
I know that in a JAVA environement it's really common to use uri to access data/files but the targeted platform is JavaScript and if i remember well i could have done nothing with the Java uri in Js.

nikki93:
According to https://github.com/marcshilling/react-native-image-picker#usage the API should put the file URI in `response.uri` rather than the path, so it should look like `content://...`. This way it is directly usable as the source URI of an image to be displayed in react-native, for example.

yfuks:
Well you get the point, a ```response.uri``` and a ```response.path``` could be a good compromise i guess, the code also need to get cleanup, i don't know if it's really clear actually. Was hopping for a real Java developer to make it "real".

Will do both and make a **PR** soon.

Thanks for your note @nikki93

PR #65, opened 2015-12-24
Add video support

pjcabrera:
Initial support for recording video or picking videos from the image picker. The code now looks for 'mediaType' to determine if it should take a 'photo' or a 'video'.

Added code to use the option key 'qualityType' to set 'high', 'medium' and 'low' image quality settings. Defaults to 'medium'.

Added code to check the option key 'imageFileType'. imageFileType === 'png' exports the photo as PNG instead of JPG. Defaults to JPG if imageFileType is not set or if set to any value other than 'png'.

Added some error reporting when some steps fail, rather than just writing to the console and either ignoring the error or returning early.

Updated the version number in package.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,19 @@
 # react-native-image-picker
 A React Native module that allows you to use the native UIImagePickerController UI to either select a photo from the device library or directly from the camera, like so:
 
-### IOS - Android
-<img title="IOS" src="https://github.com/marcshilling/react-native-image-picker/blob/master/AlertSheetImage.jpg" width="50%">
-<img title="Android" src="http://i.imgur.com/jMOLd6w.png" width="49%">
+### iOS
+** Requires iOS 8 or higher
 
-**Requires iOS 8 or higher for iOS - Requires Api 11 or higher for Android**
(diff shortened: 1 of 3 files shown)


natorojr:
+1

marcshilling:
Planning to review this soon...busy with some real job stuff at the moment :)

danlopez:
+1

natorojr:
Thanks @marcshilling

marcshilling:
Merged! 

There was an issue with the code that moved the video to the temporary directory. It appears that UIImagePickerController does that automatically, so it was deleting the video from the existing directory, and then when it tried to move it it was failing with "file does not exist". I hope I fixed that problem by only performing the move on captured videos.

I also renamed the `qualityType` option to `videoQuality`.

Finally, I changed the response object, so this will be a breaking change. The response is no longer `(didCancel, response) =>`, it's simply `(response)=>` and that object will contain the key `didCancel` if applicable.

README updates to follow...

pjcabrera:
Great! Thanks for letting me know of the API changes. I'll update the calling code at my end when I update the package.json in my project to use your latest version on npm.

natorojr:
Good stuff! Thanks @marcshilling and @pjcabrera. Looking forward to the npm package update

marcshilling:
@natorojr should be published!

PR #68, opened 2015-12-30
Correct Android path with uppercase in README.md

sulliwane:
Linux (ubuntu 14.04) is case sensitive, it won't find the folder and installation will fail.


marcshilling:
Someone had opened an issue regarding the Linux thing, so I actually just renamed the project folder to lowercase. Thank you though!

PR #85, opened 2016-01-14
add support for animated gifs

keeth:
apps (like Giphy) can save animated gifs to the camera roll, but
attempting to resize gifs in the picker destroys them, so we need
to just copy the raw data.


sgonyea:
Hmmmmm, it seems like this change is turning every photo into a GIF now?  I also can't seem to get base64 data out of my photos anymore.  Everything seems to be dumped into a file now.

keeth:
we're using this component to upload JPEG, PNG and GIF, and all seem to work fine.

PR #88, opened 2016-01-16
Handle case where we can't extract the image path in Android

pedro:
The method `getRealPathFromURI` doesn't always work with Android 4.4 and newer – try using Google Photos, for instance, and you'll see it's unable to get the path, returning a `null`. As a result all calls relying on this path are throwing an exception and causing the app to crash.

I'm really no experienced Android developer, but it seems like part of the problem is that [we can't assume the URI returned by the intent always maps to a file](http://stackoverflow.com/a/20402190/16882). For this reason I'm proposing here that we don't try to fetch image details when we don't have a local file, which suits my purposes.

The best approach would obviously be to find a way to extract a path for images coming from 4.4+ and Google Photos. They have the URI like `content://com.google.android.apps.photos.contentprovider`. [Some developers say it's possible](http://stackoverflow.com/a/20559175/16882), but after putting a few hours porting that solution (and [aFileChooser's implementation itself](https://github.com/iPaulPro/aFileChooser/blob/48d65e6649d4201407702b0390326ec9d5c9d17c/aFileChooser/src/com/ipaulpro/afilechooser/utils/FileUtils.java#L257)) to this package I just wasn't able to get it to work – still getting `null` path.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -270,6 +270,10 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
             && quality == 100) {
         response.putInt("width", initialWidth);
         response.putInt("height", initialHeight);
+    } else if (realPath == null) {
+      response.putString("error", "could not resize image");
+      mCallback.invoke(response);
+      return;
     } else {
         uri = getResizedImage(getRealPathFromURI(uri), initialWidth, initialHeight);
         realPath = getRealPathFromURI(uri);
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -236,24 +236,26 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
         return;
     }
 
-    try {
-        ExifInterface exif = new ExifInterface(realPath);
-        int orientation = exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL);
-        boolean isVertical = true ;
-        switch (orientation) {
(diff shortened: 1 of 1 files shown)


yfuks:
Hello,

first of all thanks for the **PR** !

Never heard about this issues before, like you i'm not familiar with Android, looks like Android will always surprise me.

In **< 4.4** there is closely the same issue, ```getRealPathFromURI``` return some kind of url in the case of an *external* photo (like Google Photos, Facebook Photos...) that you can make nothing with (with the actual code) so i just put a [workaround](https://github.com/marcshilling/react-native-image-picker/blob/master/android/src/main/java/com/imagepicker/ImagePickerModule.java#L231) before doing the "real joob".

From what i see from your commits you could have done a workaround much simplier ! ;)
as folow:
```java
// ImagePickerModule.java:222
String realPath = getRealPathFromURI(uri);

if (realPath == null) {
  respones.putString("uri", uri);
  response.putString("error", "could not get image path");
  mCallback.invoke(response);
  return;
}
```

Returning the uri is a good thing. In the worst case it's set uri field to null (``` source: {uri: null, isStatic: true} ``` don't put an error anymore in react-native) and in other case you can diplay it in your app (better than nothing)

Thanks for all the times you put into this work. Will investigate further soon. And thanks for the documentation !

pedro:
@yfuks hey, thanks for the awesome feedback!

Working further here I actually started handling the `null` path by using streams to create a new temp file from the `content://` url – what do you think? Not in love with creating a new file when it's obviously already on the device, but comes with the added benefit of handling any protocol the phone is capable of handling really (so no more `isUrl` check, I think). Can open a new pull if you think this is worth pursuing!

yfuks:
Just to answer quickly (it's 4:00 am in France need to rest a bit) : 
if think it's worth

pedro:
haha! cool no rush :} new pull in https://github.com/marcshilling/react-native-image-picker/pull/89

PR #89, opened 2016-01-16
Handle case where we can't extract the image path in Android (v2)

pedro:
Previous discussion: https://github.com/marcshilling/react-native-image-picker/pull/88

Here instead of just rejecting images without a path we actually create a new temporary file and copy from whatever URI was returned to the image intent. Tested this handles the `content://` URI in 4.4+.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -277,6 +277,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
     }
 
     response.putString("uri", uri.toString());
+    response.putString("urlPath", realPath);
 
     if (!noData) {
         response.putString("data", getBase64StringFromFile(realPath));
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -229,6 +229,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
       }
       catch(Exception e) {
         response.putString("error", "Could not read photo");
+        response.putString("uri", uri.toString());
         mCallback.invoke(response);
         return;
       }
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -234,21 +234,6 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
       }
     }
 
-    boolean isUrl = true;
-    try {
-        URL url = new URL(realPath);
-    }
-    catch (MalformedURLException e) {
-        isUrl = false;
-    }
-    if (isUrl) {
-        // @todo handle url as well (Facebook image, etc..)
-        response.putString("uri", uri.toString());
-        response.putString("urlPath", realPath);
-        mCallback.invoke(response);
-        return;
-    }
-
     try {
         ExifInterface exif = new ExifInterface(realPath);
         int orientation = exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL);
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -287,7 +287,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
     }
 
     response.putString("uri", uri.toString());
-    response.putString("urlPath", realPath);
+    response.putString("path", realPath);
 
     if (!noData) {
         response.putString("data", getBase64StringFromFile(realPath));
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -277,6 +275,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
     }
 
     response.putString("uri", uri.toString());
+    response.putString("urlPath", realPath);
```
pedro: While we're here should we just call this `path`?

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -220,20 +221,17 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-    try {
-        URL url = new URL(realPath);
-    }
-    catch (MalformedURLException e) {
-        isUrl = false;
-    }
-    if (isUrl) {
-        // @todo handle url as well (Facebook image, etc..)
-        response.putString("uri", uri.toString());
-        response.putString("urlPath", realPath);
```
yfuks: Why remove the fix for **< 4.4** ?
You can remove it but partially and use your fix for **< 4.4** aswell
```java
String realPath = getRealPathFromURI(uri);
boolean isUrl = true;

if (realPath !=  null) {
  // test for < 4.4
  try {
    URL url = new URL(realPath);
  } catch (MalformedURLException e) {
   isUrl = false;
  }
} else {
  isUrl = false;
}

if (realPath ==  null || isUrl) {
 try {
    File file = createFileFromURI(uri);
    realPath = file.getAbsolutePath();
    uri = Uri.fromFile(file);
  }
  catch(Exception e) {
     response.putString("error", "Could not read photo");
     mCallback.invoke(response);
     return;
   }
}
```

android/src/main/java/com/imagepicker/ImagePickerModule.java
yfuks: ```urlPath``` was just for debug, i don't think it's usefull to send it to react-native since we allready send the uri

yfuks:
Hello @pedro,

I just comment the diff. If you whant to make the folowing changes before i merge :)

android/src/main/java/com/imagepicker/ImagePickerModule.java
pedro: In my case I still needed the path to upload the image to S3. Is there a way to get a handle of the original image in Android with the uri alone?

yfuks: Are you sure about that ? Most common usage for image is to send the data in base64 ! Are you sure you don't have a module or something to convert your path into data then send it to S3 ?

pedro: Ended up writing the S3 upload in Java, didn't find any library to help.

Getting the raw image in base64 and uploading from RN could work too, but I had issues manipulating large files in RN land. Packages like [react-native-fs](https://github.com/johanneslumpe/react-native-fs) only work with path too.

yfuks: What kind of issues ?
In your module you don't convert you filePath into data ?

Every upload provider now have a rest API. You can upload from RN without any problem (never face one)
[Amazon S3 Browser-Based Upload using HTTP POST](http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-post-example.html)

There is an exemple (with cloudinary rest API)
https://github.com/marcshilling/react-native-image-picker/issues/61#issuecomment-164467359

android/src/main/java/com/imagepicker/ImagePickerModule.java
pedro: oh good call! Updated

android/src/main/java/com/imagepicker/ImagePickerModule.java
pedro: > What kind of issues ?

App was crashing with large photos, if I remember correctly there's some sort of limit on how many bytes you can allocate on a RN string..?

> In your module you don't convert you filePath into data ?

I don't – that's why I ended up writing the upload in Java, eek :} My `@ReactMethod` in Java just receives the S3 url and the file path, and uploads the corresponding file using `HttpURLConnection` and a buffer.

yfuks: > there's some sort of limit on how many bytes you can allocate on a RN string

:/

That's quite something to deal with.

Well since the issues could block other people i guess returning path to **RN** will help and it don't costs too mutch.

pedro: Right on :\ but perfect, updated and `path` it is. Thanks!

yfuks:
Awesome work @pedro !
Just try it and it work perfectly for **< 4.4**

pedro:
phew, great to hear! thanks a lot for all the help.

yfuks:
@pedro 
> App was crashing with large photos

Just find that the module was crashing with large photo because bitmap ```scaledphoto``` was not recycle

fix in **0.13.5**

https://github.com/marcshilling/react-native-image-picker/commit/ff09909727040b4f1910b4fed5e6381798308ff1

yfuks:
Now picker can crash in only one case =>
If the image is > 16MB

> Android devices can have as little as 16MB of memory available to a single application.

documentation : 
http://developer.android.com/training/displaying-bitmaps/index.html

pedro:
@yfuks a-ha, amazing! Going to give it a try, thank you!

PR #91, opened 2016-01-18
handle ActivityNotFoundException

pedro:
When starting intents to get a photo from the camera or photo album.


yfuks:
Robustness code allways good ! Nice one @pedro !

PR #93, opened 2016-01-18
Resize tweaks

pedro:
- Use a temp file for the resized images. Otherwise they are displayed under the "pictures" folder within my app files
- Fix path for resized images, make sure it points to the resized file

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -299,8 +299,9 @@ public class ImagePickerModule extends ReactContextBaseJavaModule {
         response.putInt("width", initialWidth);
         response.putInt("height", initialHeight);
     } else {
-        uri = getResizedImage(getRealPathFromURI(uri), initialWidth, initialHeight);
-        realPath = getRealPathFromURI(uri);
+        File resized = getResizedImage(getRealPathFromURI(uri), initialWidth, initialHeight);
+        realPath = resized.getAbsolutePath();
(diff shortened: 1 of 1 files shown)


yfuks:
Well puting it in the pictures folder was a choice, you should look at the storageOptions that are implemented in **iOS**

```javascript
storageOptions: { // if this key is provided, the image will get saved in the documents directory (rather than a temporary directory)
    skipBackup: true, // image will NOT be backed up to icloud
    path: 'images' // will save image at /Documents/images rather than the root
}
```

Make the same API for android and iOS sure is beneficial. Since iOS was there first try to match it as mutch as possible.

Adding the options to make the user choose will probably do the job.

pedro:
ah interesting! I didn't realize the storageOptions applied to resized images, makes total sense.

Can I add support for a temporary path then? Our use case is to allow users to pick an image from their album, and not have that show anywhere else.

pedro:
eg:

```js
storageOptions: {
  tmp: true // save resized images in a temporary location
}
```

yfuks:
Would be great !

```javascript
storageOptions: {} // if this key is provided, the image will get saved in the documents/picture directory (rather than a temporary directory)
```

Could be good too ! The choice is up to you ! :)

pedro:
ah, doh! missed that :} ok updated to infer from the presence of `storageOptions`.

Thanks!

yfuks:
Sorry for the wait, will merge, add some features and publish a new version

pedro:
no worries! thank you :}

PR #105, opened 2016-02-11
Fix iOS7 crash

umfana:
containsString is an iOS8+ method and as such crashes on iOS7. Using
rangeOfString fixes that.


marcshilling:
Published in version 0.13.6...thanks!

PR #123, opened 2016-03-01
Uri for video returns real path

Anthonyzou:
This seems more useful than getting a content uri.


yfuks:
What do you mean by 
> this seems

An uri : 
> A Uniform Resource Identifier that identifies an abstract or physical resource, as specified by RFC 2396.

Anthonyzou:
The problem is the return Uri for picking a video is something like ` { uri: 'content://media/external/video/media/30267' }`. This isn't useful and not an actual file path.

marcshilling:
@Anthonyzou since it seems you are the only one who has asked for this feature, I'd recommend just forking the repo...simple solution.

PR #128, opened 2016-03-03
Invoke response callbacks after UIImagePickerController dismiss animation has completed

pjcabrera:
In my app, I am displaying a React Native Alert in the response callback from RNImagePicker. This works fine on iPhone and iPod touch. But on iPad, the Alert doesn't show, and I get a warning in the Xcode console:

`Warning: Attempt to present <UIAlertController: 0x7ad6ba00> on <UIImagePickerController: 0x7ac63c00> whose view is not in the window hierarchy!`

I can reproduce this in iPads with different models and in the iPad simulator. The Alert doesn't show and this warning only occurs if you display an React Native Alert in the response callback in iOS on iPad. Everything works as intended on iPhone & iPod touch devices.

I believe what is happening is that the React Native Alert iOS implementation sets a reference to the UIImagePickerController as its popoverViewController, but this UIImagePickerController is not part of the RN view hierarchy, which leads to the iOS warning and the Alert not showing.

The fix is to make sure the response callback is only invoked after the UIImagePickerController dismiss animation has completed. With the UIImagePickerController gone, the React Native Alert iOS implementation uses the RN RootViewController as the popoverViewController.

I tested this PR on different model iPads and iPhones, on iOS 8.x and iOS 9.x, and the expected result now occurs on both device types without warnings in the Xcode console. Also, response callbacks without use of Alerts continue to work as before.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -247,178 +247,180 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
 
 - (void)imagePickerController:(UIImagePickerController *)picker didFinishPickingMediaWithInfo:(NSDictionary *)info
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [picker dismissViewControllerAnimated:YES completion:nil];
-    });
-    
-    NSURL *imageURL = [info valueForKey:UIImagePickerControllerReferenceURL];
-    NSString *mediaType = [info objectForKey:UIImagePickerControllerMediaType];
-    
-    
-    NSString *fileName;
-    if ([mediaType isEqualToString:(NSString *)kUTTypeImage]) {
-        NSString *tempFileName = [[NSUUID UUID] UUIDString];
-        if (imageURL && [[imageURL absoluteString] rangeOfString:@"ext=GIF"].location != NSNotFound) {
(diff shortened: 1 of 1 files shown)


marcshilling:
Is this just to fix an Xcode warning or is something actually broken for you?

pjcabrera:
It's not the warning that matters, but the React Native Alert in the callback doesn't show on iPad.

I tried different solutions, such as setTimeout in the callback, but these weren't addressing the root cause, they were more like workarounds.

pjcabrera:
Again, the issue is that if I create a RN Alert in the response callback from RNImagePicker, they don't show on iPad, and a runtime iOS warning shows in the console to let the developer know why.

Let me know if you need an example project to try to reproduce the issue, or if you have any workaround suggestions.

marcshilling:
Gotcha. I'm on my phone now and the diff looks pretty crazy, so I'll review when I'm back on desktop...probably tomorrow.

pjcabrera:
No worries, I don't expect you to accept PRs without proper review, and certainly not from just looking at it on your phone!

marcshilling:
Ah, the diff only looked so crazy because the extra indent level inside the completion block! Good work and thanks!

pjcabrera:
Thanks for reviewing and releasing a new npm package so quickly!

Issue #121, opened 2016-02-29
Android no picture back


rainmeterLotus:
I run on android,use "ImagePickerManager.launchImageLibrary" ,but nothing back!!!the log is not !!

rainmeterLotus:
because of ImpagePickerMoudle's onActivityResult no call

marcshilling:
The API recently changed...update to the latest version, update your code based on the README, and then if you are still having issues come back and post some of your code.

marcshilling:
@rainmeterLotus any update here?

agutoli:
Hello @marcshilling @rainmeterLotus, I fix it and solve for me... https://github.com/marcshilling/react-native-image-picker/pull/130

marcshilling:
Closing because this is a React Native versioning issue. This library will be kept up to date with the latest version of RN.

PR #130, opened 2016-03-06
fix issue #121 (Android no picture back)

agutoli:


marcshilling:
It seems like this is just undo-ing an update we made recently and I don't think this is correct. 
@yfuks can you review?

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -257,12 +257,13 @@ public void launchImageLibrary(final ReadableMap options, final Callback callbac
     mCallback = callback;
 
     try {
-      currentActivity.startActivityForResult(libraryIntent, requestCode);
+      mReactContext.startActivityForResult(libraryIntent, requestCode, null);
```
yfuks: startActivityForResult from reactContext ? Really bad pattern.

yfuks:
Yes @marcshilling like you said.

The **PR** just make an example of how to install the module for **RN <= 0.17.0** and remove it for current version.

It's a good update to the readme but you don't need to remove the current version @agutoli  :)

agutoli:
@yfuks Tks for the feedback... I will try to refactor this and send a new pull request soon. Hug!

PR #133, opened 2016-03-07
customButtons for android fix

JeJan:
add customButtons support for android


yfuks:
We are using the AlertDialog in **Android**

> A dialog that can show a title, up to three buttons, a list of selectable items, or a custom layout. 
http://developer.android.com/guide/topics/ui/dialogs.html

Are you sure it's working even when you have a photo and a library button ?

JeJan:
AlertDialog has support for "items" as list, for me this works nice

yfuks:
I choose #134 implementation since it was a bit more clear !
Still a great **PR** !

PR #143, opened 2016-03-09
reorient image on image resize

jeremyeaton89:
When images are too large on our nicer Android devices, our image orientation is incorrect. From what I learned from [this article](http://voidcanvas.com/whatsapp-like-image-compression-in-android/), the correct orientation is restored, and the `angle` option is still regarded. 

I tested on both a phone that had this issue and one that didn't. 

Hopefully the code is sound - I don't code in Java, really.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -501,6 +501,23 @@ public class ImagePickerModule extends ReactContextBaseJavaModule implements Act
     matrix.postRotate(angle);
     matrix.postScale((float) ratio, (float) ratio);
 
+    ExifInterface exif;
+    try {
+      exif = new ExifInterface(realPath);
+
+      int orientation = exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, 0);
+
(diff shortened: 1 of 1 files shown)


yfuks:
A great **PR** but the re-orientation will only appear if we resize the given image.
And it nuliffy the ```Angle``` option.

jeremyeaton89:
In our use case, we pass the `maxWidth` and `maxHeight` options, which resizes the image, and, if I understand correctly, reorients the image. 

Am i, perhaps, using the library wrong? I noticed you handled some reorientation cases when upgrading from `0.18.0`, but `0.18.3` didn't seem to help our case.

jeremyeaton89:
>And it nuliffy the Angle option.

I don't believe this is the case because the code is adding an extra rotation based on the orientation of the phone, not the image, if i'm understanding correctly. My testing also shows that image adheres to `angle` as one would expect.

Perhaps a better name for this PR would be "Reorient image on image resize"

yfuks:
Well it work perfectly ! I guess i was wrong about the ```angle```. I think it could be good too if we can reorient the image even if the user don't resize/set quality/set angle.

PR #147, opened 2016-03-12
Android private files

markomatic:
* Ability to save images into android private files directory `/data/data/<APP_PACKAGE>/files/images` if `storageOptions.savePrivate`
* Android code refactored and reformatted to be more human readable


yfuks:
> Android code refactored and reformatted to be more human readable

I don't really see the diff. 4 space for indentation instead of 2 isn't more readable..

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -13,27 +13,12 @@
...
-
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+import java.io.*;
```
yfuks: Are you really sure that importing all **java.io** is a good things to do ?

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -42,566 +27,714 @@
...
+            currentActivity.startActivityForResult(libraryIntent, requestCode);
+        } catch (ActivityNotFoundException e) {
+            e.printStackTrace();
+        }
     }
 
-    titles.add(cancelButtonTitle);
-    actions.add("cancel");
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        if (isValidResult(requestCode)) {
```
yfuks: if it's valid we return ?

android/src/main/java/com/imagepicker/ImagePickerModule.java
markomatic: That was rhetorical question :smile: Fixed it...

android/src/main/java/com/imagepicker/ImagePickerModule.java
markomatic: My big mistake... Function body was doing something opposite of its' name... thnx

markomatic:
@yfuks you are welcome :smile: I needed picker result to be in apps private files. I extracted some code into methods and made some changes in private variable names. 4 spaces for indentation is de-facto standard, so it's probably preferred...
To see PR without white spaces you can use `?w=1` at the end of url, [PR - no white space](https://github.com/marcshilling/react-native-image-picker/pull/147/files?w=1). It will be readable during review. Thnx.

marcshilling:
@markomatic if we're going to be changing the Android `storageOptions` behavior, I would love to get it in parity with iOS. By that I mean if `storageOptions` is provided, we save the file in the private app directory (`/data/data/<APP_PACKAGE>/files/`). If `storageOptions.path` is provided, with `path` being something like `images/test`, we would save it under `/data/data/<APP_PACKAGE>/files/images/test`. If you want to make that change and create a new PR, I would accept it. Please leave the "Android code refactored and reformatted to be more human readable" out of it, or submit a separate PR for that change alone.

markomatic:
@marcshilling, great. I will open new pull request in next few days when I get time. I will close this PR now.

PR #160, opened 2016-04-01
Add image file size to the response

aksonov:
Add image file size to the response


marcshilling:
Good stuff! I changed "size" to "fileSize" for clarity.

PR #163, opened 2016-04-06
Add duration limit option to iOS

### LINKED ISSUE #115: [iOS] Video Recording Max Time
I think it would be nice if we could limit the video recording max time

barodeur:
Hello,

Here is a very simple implementation of the `durationLimit` option for iOS using [`picker.videoMaximumDuration`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIImagePickerController_Class/#//apple_ref/occ/instp/UIImagePickerController/videoMaximumDuration)

This would probably close #115.


yamill:
sweet! thanks, i'll try this out.

PR #173, opened 2016-04-15
Android - storage access framework

mlabrum:
Hey,

This is my first time submitting a pull request for a native module, so hopefully this is fine :)

This updates the image picker to use http://developer.android.com/guide/topics/providers/document-provider.html#client and also allows the user to specify a mimetype to filter the list.
(this change only affects the image chooser)

I've also added 
openFileUri -> opens a file uri returned by the picked image so the user can open that file at a later point in time
readUri -> reads the file URI and returns the content in base64 format

Let me know if you need any changes done :)

(a college of mine is working on doing the same for IOS....)

Cheers,
Matt

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -337,6 +337,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule implements Act
       realPath = getRealPathFromURI(uri);
     }else if(response.hasKey("path")){
       realPath = response.getString("path");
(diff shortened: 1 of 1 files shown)


marcshilling:
Correct me if I'm wrong (I'm just viewing the code, haven't actually built it yet), but it seems you killed off the `allowEditing` option when choosing from library?

yfuks:
Yep the ```allowsEditing``` options don't work anymore when choosing from library.
And that just a personnal vision but that's a bit out of module purpose, since the module is called "image picker"

Since there is no response i close the **PR**.

PR #176, opened 2016-04-17
Add ref to asset library version for color grabber

chetstone:
Add origURL to the returned object. This is a reference to the asset library handle of the original image. It can be passed to [react-native-color-grabber](https://github.com/bsudekum/react-native-color-grabber)


marcshilling:
Looks good to me @yfuks. I think we definitely need this functionality.

I'll let you merge and publish if you're ok with it!

chetstone:
I just thought of something – I was only using this component to pick images from the library. So I haven't tested this fix with getting images from the camera directly... Don't know if it makes any difference, I suppose the camera image is added to the asset library before this component processes it.

PR #200, opened 2016-05-10
Show all media types

rahuljiresal:
When the media types are not specified while calling the API, show all of them.


marcshilling:
Not so simple because now you're removing the ability for a person so see only images. This is react-native-**image-picker**, it should default to images only. If you want to add a different `mediaType` such as `mixed`, for the user to explicitly choose to show all media, that might work. But I am concerned about the response object. This will need to be tested thoroughly for all media types.

qimingweng:
Agreed that a `mixed` option is probably more useful

marcshilling:
No activity here, so closing this. Feel free to send another PR with a `mixed` option instead.

PR #222, opened 2016-06-10
master

npomfret:
Updated some android meta data - changed type of gps coords, and added a UTC timestamp


marcshilling:
In my opinion - lat/lon should not be included in the response if they are `0`. Also, including both `timestamp` AND `timestamp_epoch_millis` is a little redundant. Why not just return `timestamp` in correct UTC format? Make these changes and I'll merge. It would also be great if you could update the README to reflect the new response items...but I can do that if you don't want to.

npomfret:
Hey - thanks for the response.  The only reason for keeping the timestamp field is that it's human readable, but yes - totally redundant.

I'll make the changes.

npomfret:
https://github.com/marcshilling/react-native-image-picker/pull/223

PR #223, opened 2016-06-10
master

npomfret:
* Changed the type of the gps data to be numbers (if they are present)
* Change timestamp field to be millis since 1970

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -40,8 +40,13 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.text.ParsePosition;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
+import java.util.Locale;
+import java.util.TimeZone;
 import java.util.UUID;
 
 public class ImagePickerModule extends ReactContextBaseJavaModule implements ActivityEventListener {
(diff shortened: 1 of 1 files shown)


marcshilling:
Looks good. Code appears correct...any idea why a photo I just took on a Samsung Galaxy S6 wouldn't have lat/lon? Not super familiar with Android phones...but I have location services enabled.

marcshilling:
Ah, nevermind. I had location disabled under camera settings. Looks good! Merging shortly

marcshilling:
@npomfret hang on...what is this time format? A photo I just took returns `timestamp: 977867064`. That doesn't seem correct in any form of epoch...

marcshilling:
I'm fine with epoch seconds/mills if they are correct but ISO format might be our best bet: `2014-09-08T08:02:17-05:00`

npomfret:
Yes, that doesn't seem right at all.  I copied the code from direct from a recent copy of the android source code (`ExifInterface.getDateTime`).  What was the text representation before if got converted?

npomfret:
Hmm, can't see annoying obviously wrong with the code - and it works perfectly on my device. I guess its possible that different devices save the time in different formats??

marcshilling:
Android's documentation is pretty pathetic....any idea the difference between `SubSecTime`, `SubSecTimeDig`, `SubSecTimeOrig`?

marcshilling:
I'm not sure I even understand the need for the `subSecs` at all...I'm going to make some changes myself and return it as an ISO string

marcshilling:
Cool, done. `timestamp` now returns in ISO UTC format like `2016-06-10T16:28:32Z`. Thanks for your work on this!

PR #245, opened 2016-06-24
Upading the readme, to stop others from wasting time on a known andro…

SudoPlz:
I hope this helps others.


yfuks:
That's not an issue with this module.
Since we don't even use ```onActivityResult``` method.

PR #258, opened 2016-07-08
Set source based on Platform

phoenixbox:
I thought the inclusion of a Platform check in the sample code might be helpful to those new to react-native

Thanks for the lib ❤️ 🚀 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ public class MainActivity extends ReactActivity {
 ## Usage
 
 ```javascript
+var Platform = require('react-native').Platform;
 var ImagePicker = require('react-native-image-picker');
 
 // More info on all the options is below in the README...just some common use cases shown here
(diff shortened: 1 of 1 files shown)


jakecraige:
Seeing this now, is there a chance this is actually confusing to people since it doesn't seem to be required to strip the file prefix? (at least I don't think so, I'm not an it works)

marcshilling:
@jakecraige that part of the README is a little dated. If you can confirm that the same `uri` works on both iOS and Android I suppose we can remove it.

phoenixbox:
Hmmm not sure, came across this a while ago so Im not 💯 that the url will work in Android without stripping the file prefix.

I mainly made this update because it was confusing to me to see `const source` be redefined without a conditional Platform check. If the url works without the prefix, then no worries rolling back this commit and settling on a single line 🌈

marcshilling:
I can test it out when I get some time.

phoenixbox:
awesome 🙌 , thanks for the lib 🌮  super helpful ✨

jakecraige:
Sounds great. It's been working well for me without it which Is why I brought this up. Definitely test for yourself to make sure I'm not crazy thought 😄

marcshilling:
Just confirmed @jakecraige, good call. Updating the README/Example project now!

PR #262, opened 2016-07-18
modify for jzfp

p2mThinknet:
modify for jzfp


marcshilling:
Can you explain what you're doing here? I don't quite understand it.

yfuks:
no response

Issue #252, opened 2016-07-04
[iOS] Saving photos into the default camera roll directory


vessp:
Hello, in my app I am using the camera to take pictures by calling **ImagePicker.launchCamera** however I would like the pictures to be saved into the default ios camera roll directory so the user has easy access to them later if they would like.  

I'm pretty new to ios so I don't really know the file structure very well.  I'm not sure if this path argument is relative to the root directory of the filesystem or the root directory of the app folder, but could i do this by setting the **storageOptions.path** to something like **/private/var/mobile/Media/DCIM/**?  

Thanks!

marcshilling:
This isn't supported right now. We would need to add a call to [UIImageWriteToSavedPhotosAlbum](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKitFunctionReference/index.html#//apple_ref/doc/uid/TP40006894-CH3-SW29) somewhere in the code

gilesvangruisen:
Opened a PR #281 to add this functionality on iOS

marcshilling:
This is live in `v0.21.5` - thanks @gilesvangruisen!

PR #281, opened 2016-08-04
Save to camera roll option

gilesvangruisen:
This adds an option under `storageOptions` called `cameraRoll` which, if set to `true`, will save the resulting cropped image to the iOS Camera Roll.


gilesvangruisen:
This should solve #252

gilesvangruisen:
This doesn't yet support video. I should add that

marcshilling:
Good stuff! I'll review this shortly.

mozilla-rowvigor:
Is there any plans to insert the camera roll option to support video?

PR #301, opened 2016-09-05
add imageSource from photo or library

yangjiayu:


yfuks:
I'm not sure that's a correct things to do.
Calling showImagePicker and put an option to "not show imagePicker" seems a bit confusing.

The purpose of this **PR** (i guess) is to choose directly between camera and library but this can be done with 
```javascript
// Launch Camera:
ImagePicker.launchCamera(options, (response)  => {
  // Same code as in above section!
});

// Open Image Library:
ImagePicker.launchImageLibrary(options, (response)  => {
  // Same code as in above section!
});
```

PR #303, opened 2016-09-09
Support RN 0.33.0

ramilushev:
Be sure to merge this after 0.33.0 is released

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -40,5 +40,5 @@ repositories {
 }
 
 dependencies {
-    compile 'com.facebook.react:react-native:0.20.+'
+    compile 'com.facebook.react:react-native:+'
 }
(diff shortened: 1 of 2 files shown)


marcshilling:
@ramilushev any idea how to make this backwards compatible? I merged the change, but now that breaks this library for any users not on `v0.33`. I'm going to revert the changes for now.

ramilushev:
@marcshilling I don't think there is a way around it .. we could add a note in the documentation, like in [react-native-fcm](https://github.com/evollu/react-native-fcm#note), that users should use <(some version) for RN <0.33.0

marcshilling:
@ramilushev I figured out a way to do it...it's live on npm `v0.22.5`...thanks for your help!

PR #320, opened 2016-09-20
[ios] Fixing support to png

gullitmiranda:


marcshilling:
What's wrong with the existing code?

gullitmiranda:
`[self.options objectForKey:@"imageFileType"]` is already an `NSString`.

stack: 
- react-native v0.33.0
- xCode 7.3.1
- iOS 9.3

marcshilling:
Ok sure, but `objectForKey` returns an `NSObject`, not an `NSString`. Does the extra cast really matter? Was anything not working before or is this simply a syntax change?

marcshilling:
No response.

aksonov:
I'm getting crash 
```[NSTaggedPointerString stringValue]: unrecognized selector sent to instance 0xa00000000676e703```

when set 
```
    imageFileType: 'png',
```

what am I doing wrong?

marcshilling:
Where are you setting this code?
`imageFileType: 'png',`

aksonov:
Yes, all I want is to avoid any compression/modification of selected image.

aksonov:
I'm trying to set it via options JS object

aksonov:
@marcshilling Any chance that you could apply this PR to fix the problem?

marcshilling:
I'll have to look into it. If you're just looking to avoid compression, if you specify `quality: 1` then it should not be compressed, but it will be a `.jpg` file.

aksonov:
"f you're just looking to avoid compression"
@marcshilling It is not true. Image with quality: 1 is still processing and it results in much much large size than original (4x times!). Is it possible to get original image? I've tried to use 'originalUri' but it seems not accessible (starting with assets://) from React Native.

nica0012:
@marcshilling Hi Marc, how come when adding a png image with a transparent background it is returned with a white background? Any way around this?

PR #370, opened 2016-10-25
add license key to package.json file and fixed a crash when use imageFileType

tangkunyin:
`In fact,when I use imageFileType of png,crash happend below:`

crashed


yfuks:
Don't know about the **iOS** part but on **Android** you basically reverse the 20 last commit from this module...

marcshilling:
Yeah there's too many files you've changed here for no reason, like stuff in the Example project. If you want to clean this up and submit a new one with just the proposed change that would be great, because I don't really understand what's going on here. Also, please don't change the iOS file indentation - it makes for a ton of diffs when looking at the change.

PR #387, opened 2016-11-10
fix compile problem

npomfret:
fix compile problem


marcshilling:
This undoes [all the work](https://github.com/marcshilling/react-native-image-picker/commit/05309a82c547cb86b566003aa3c24bfc8afd0fd0) that was done to support RN 0.33+

npomfret:
It should still work - both versions of the `onActivityResult` methods are still implemented

Issue #199, opened 2016-05-09
[Android] Incorrect dimensions / orientation for image from library


thomasjonas:
When I use ```ImagePickerManager.launchImageLibrary``` on my Android device (Sony Ericsson LT26i) and select a portrait image it returns the following response:

   ```
{ 
     type: 'image/jpeg',
     fileName: 'IMG_20160502_182352.jpg',
     fileSize: 1657747,
     path: '/storage/sdcard0/Pictures/RCTCameraModule/IMG_20160502_182352.jpg',
     uri: 'content://media/external/images/media/1937',
     height: 3000,
     width: 4000,
     isVertical: false
}
```

So according to the response we're dealing with a landscape image, but that's incorrect. When I log the orientation (using [this line of code](https://github.com/marcshilling/react-native-image-picker/blob/4c6abc024a19dcb06267d5ffb7907375685196ef/android/src/main/java/com/imagepicker/ImagePickerModule.java#L379)) it says the image has this orientation: ORIENTATION_ROTATE_90.

Doesn't that mean that the 4000x3000 (width x height) should be rotated 90 degrees to 3000 x 4000 (width x height) and that isVertical: true? According to [this link](http://www.daveperrett.com/articles/2012/07/28/exif-orientation-handling-is-a-ghetto/) handling image rotation / exif data is not that easy as it looks like...

thomasjonas:
I see that the CameraRoll component from react-native also has this problem. I think using just `BitmapFactory.decodeFile` to determine the orientation and dimensions is probably not enough.

yfuks:
Can you try with more exemples and tell me if you always encounter the issue ?
(Just watch the code and i feel like ```isVertical: false``` and ```isVertical: true``` are maybe inverted)

marcshilling:
Old issue + no response from OP = closing

Vietvantue:
I also face the same issue at latest version

PR #388, opened 2016-11-10
Send originalRotation in response

jemise111:
Hey @marcshilling, awesome library!

This PR addresses #199 (issue was closed but not fixed).

For some devices (notably Samsung Galaxy S3, S5) and image will have a rotate value (90 deg) in the Exif data even if the image was never rotated.

So if you try to upload the image somewhere and then later grab it and render it, the image is rendered in the wrong orientation compared to the original.

I'm not sure if this needs to be handled at the library level but this PR simply at least passes that value as part of the response so if an image has a rotate value attached to it the client can decide how to handle (like transform the image 90deg to account for it)

Note: this hasn't been an issue for iOS but I figured I'd add it, I can remove it if that's easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,4 +209,4 @@ path | - | OK | The file path
 latitude | - | OK | Latitude metadata, if available
 longitude | - | OK | Longitude metadata, if available
 timestamp | - | OK | Timestamp metadata, if available, in ISO8601 UTC format
-originalRotation | OK | OK | Rotation degrees (photos only) *See [#109](/../../issues/199)*
+originalRotation | - | OK | Rotation degrees (photos only) *See [#109](/../../issues/199)*
--- a/README.md
+++ b/README.md
@@ -209,3 +209,4 @@ path | - | OK | The file path
 latitude | - | OK | Latitude metadata, if available
 longitude | - | OK | Longitude metadata, if available
 timestamp | - | OK | Timestamp metadata, if available, in ISO8601 UTC format
+originalRotation | OK | OK | Rotation degrees (photos only) *See [#109](/../../issues/199)*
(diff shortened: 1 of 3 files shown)


malitta:
I just tested this branch after running in to issue #199 myself, seeing there is no other alternative. And it works perfectly! Hope this get's merged.

marcshilling:
@jemise111 if this isn't an issue for iOS I'd prefer we only add it on Android. Can you update?

jemise111:
@marcshilling Done!

Romchik22:
@jemise111  Unfortunately after converting image to base64 and reconvert back, Image component not understand in which rotation should showing image. What do you think about this?

jemise111:
@Romchik22 I'm a little confused here. Despite whether you use the base64 data off the response or not you should still be able to look at `response.originalRotation` and know whether or not that image needs to be rotated, right?

Romchik22:
@jemise111 Sorry for my English. You are right. My problem is after uploading image on server and loading it back image lost orientation value and displaying incorrect.  I have solved this problem. Now I firstly rotate image and after this convert to base64.

jemise111:
@Romchik22 👍 glad to hear it

angusmccloud:
@jemise111 Do you have a working example of how to implement this fix? 

For instance if I run the below code I get an error when trying to alert the rotation.  Is there something I need to do differently to grab the rotation (and if you have it, what's the line of code you run to rotate the image). 

I appreciate the help if you have something that works!

ImagePicker.showImagePicker(options, (response) => {
      if (response.didCancel) {
        console.log('User cancelled image picker');
      }
      else if (response.error) {
        console.log('ImagePicker Error: ', response.error);
      }
      else if (response.customButton) {
        console.log('User tapped custom button: ', response.customButton);
      }
      else {
//        Alert.alert(response.originalRotation);

PR #394, opened 2016-11-22
Add fileName, latitude, longitude and timestamp response fields for iOS

interstateone:
This PR adds response fields for `fileName`, `latitude`, `longitude` and `timestamp` values on iOS. There are some caveats about when these fields are available in the response:

- `fileName` is always available for picked media and is available for media captured with the camera when `cameraRoll` and `waitUntilSaved` are both true (i.e. it needs to be in the filesystem to have a file name).
- `latitude` and `longitude` are available for picked media. Captured media doesn't get this metadata automatically, so some CoreLocation code can be added in the future to have this data for all cases.
- `timestamp` is available for picked media and is available for media captured with the camera when `cameraRoll` and `waitUntilSaved` are both true. Some more logic can be added in the future to get this for all cases.

There is still room for improvement here but this is what I needed to change for our uses, and hopefully it can be merged in order for other people to take advantage of.


interstateone:
This PR has been squashed into one commit and rebased on master to resolve merge conflicts in the readme, it should be ready for review and merging now.

PR #410, opened 2016-12-01
Update iOS RN header paths

janicduplessis:
https://github.com/facebook/react-native/commit/e1577df1fd70049ce7f288f91f6e2b18d512ff4d moves header imports to `<React/*>`. This updates header include path and all imports so build works against react master (soon to be version 0.40.0).

TODO: Update examples when 0.40 is out.


marcshilling:
This is a breaking change, yeah? So once we merge this, people will need to be using React Native >= 0.40?

janicduplessis:
Yea unfortunately this is not backward compatible so we'll have to do that.

PR #430, opened 2016-12-24
Update ImagePickerActivityEventListener to use newer React Native API.

kanerogers:
Hi, there!

This pull request fixes an issue I faced using React Native 0.39 on Android:

```
/Users/kanerogers/Development/traxi/traxi/node_modules/react-native-image-picker/android/src/main/java/com/imagepicker/ImagePickerActivityEventListener.java:6: error: cannot find symbol
import com.facebook.react.bridge.BaseActivityEventListener;
                                ^
  symbol:   class BaseActivityEventListener
  location: package com.facebook.react.bridge
/Users/kanerogers/Development/traxi/traxi/node_modules/react-native-image-picker/android/src/main/java/com/imagepicker/ImagePickerActivityEventListener.java:10: error: cannot find symbol
public class ImagePickerActivityEventListener extends BaseActivityEventListener {
                                                      ^
  symbol: class BaseActivityEventListener
/Users/kanerogers/Development/traxi/traxi/node_modules/react-native-image-picker/android/src/main/java/com/imagepicker/ImagePickerActivityEventListener.java:15: error: incompatible types: ImagePickerActivityEventListener cannot be converted to ActivityEventListener
        reactContext.addActivityEventListener(this);
```

Thanks again for a fantastic library! ❤️

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerActivityEventListener.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerActivityEventListener.java
@@ -3,11 +3,11 @@ package com.imagepicker;
 import android.app.Activity;
 import android.content.Intent;
 
-import com.facebook.react.bridge.BaseActivityEventListener;
+import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.ReactApplicationContext;
 
 
-public class ImagePickerActivityEventListener extends BaseActivityEventListener {
+public class ImagePickerActivityEventListener implements ActivityEventListener {
 
     private ActivityResultInterface mCallback;
 


marcshilling:
I do not believe this is the correct solution and you're not the first person to send a PR like this. The issue is for now, I want to support RN versions less than 0.33. See [here](https://github.com/marcshilling/react-native-image-picker/issues/324#issuecomment-248916795) for more info and a possible fix for you.

Either way, when RN 0.40 is released we we likely need to set the minimum required to that (because of #410), so we can make this issue go away then.

kanerogers:
Hi, Marc,

Not a a problem - your call! I'll continue using my fork in the interim.

Thanks again,
Kane

marcshilling:
@kanerogers this should no longer be an issue with `v0.25.0` which requires React Native 0.40.

kanerogers:
Sweet! 👍

PR #462, opened 2017-01-30
added null check for buttons title in options

philipshurpik:
Hi!

Thanks for your great library!

We found one strange bug which occur if 
`takePhotoButtonTitle == null` and 
`chooseFromLibraryButtonTitle` is less than 10 symbols plus is assigned dynamically (from texts system)

The reason is difference in serialization of strings with different length.
In details explained here:
http://stackoverflow.com/questions/33105750/nsjsonserialization-generating-nscfstring-and-nstaggedpointerstring

To solve this we added explicit check that `takePhotoButtonTitle` and `chooseFromLibraryButtonTitle` are not null

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -117,10 +117,13 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
         }
     }
 
-    if ([action.title isEqualToString:[self.options valueForKey:@"takePhotoButtonTitle"]]) { // Take photo
+    NSString *takePhotoButtonTitle = [self.options valueForKey:@"takePhotoButtonTitle"];
+    NSString *chooseFromLibraryButtonTitle = [self.options valueForKey:@"chooseFromLibraryButtonTitle"];
+
+    if (![takePhotoButtonTitle isEqual:[NSNull null]] && [action.title isEqualToString:takePhotoButtonTitle]) { // Take photo
         [self launchImagePicker:RNImagePickerTargetCamera];
     }
-    else if ([action.title isEqualToString:[self.options valueForKey:@"chooseFromLibraryButtonTitle"]]) { // Choose from library
+    else if (![chooseFromLibraryButtonTitle isEqual:[NSNull null]] && [action.title isEqualToString:chooseFromLibraryButtonTitle]) { // Choose from library
         [self launchImagePicker:RNImagePickerTargetLibrarySingleImage];
     }
 }


marcshilling:
This seems fine, but why would `takePhotoButtonTitle` or `chooseFromLibraryButtonTitle` ever be null unless the user explicitly set them to null? We give them default values in the JavaScript module. With your fix, if the user did explicitly set them to null nothing would happen when clicked. I'm assuming you are just trying to prevent a crash but I think it's safe to assume that these should just never be null because it would provide a bad experience anyway.

philipshurpik:
User can disable `takePhotoButton` if set `takePhotoButtonTitle=null`in options.

That's what we are doing (because we need possibility to choose picture only from Gallery and not with camera) 

And we set it to null - so this action will not be shown (it will not add this UIAlertAction  in showImagePicker):
```
if (![takePhotoButtonTitle isEqual:[NSNull null]] && takePhotoButtonTitle.length > 0) {
        UIAlertAction *takePhotoAction = [UIAlertAction actionWithTitle:takePhotoButtonTitle style:UIAlertActionStyleDefault handler:^(UIAlertAction * action) {
            [self actionHandler:action];
        }];
        [self.alertController addAction:takePhotoAction];
    }
```

marcshilling:
Well it sounds like you don't need the picker at all! If you only want to choose from gallery you should build your own action button and then use `ImagePicker.launchImageLibrary(options, (response)  => {});` directly.

philipshurpik:
Yeah, thanks :)
That's a good idea

But I think we need to add this check into code. Otherwise, other users can face the same problem because:
1)  In `showImagePicker` method we are checking `![takePhotoButtonTitle isEqual:[NSNull null]]`. 
It can be useful for example if you need to generate ImagePicker buttons dynamically (and sometimes pass null to not show button).

2) But in `actionHandler` such check was not performed. So it could lead to uncaught exceptions in Native code and crash of the app without a possibility to catch it from JS code.
Because when we check `isEqualToString: someString` and this `someString` has the type of NSTaggedPointerString and is null - we receive an uncaught exception.

marcshilling:
"It can be useful for example if you need to generate ImagePicker buttons dynamically (and sometimes pass null to not show button"

I disagree. No one should be passing `null`. Like I said, if people do not want to be showing certain buttons then they should not be calling `showImagePicker`. The `launchImageLibrary` and `launchCamera` convenience methods were provided for exact that reason - you can build the UI however you'd like and then access the underlying functionality.

PR #464, opened 2017-02-01
Added support of taking photos for Android >=7.0

rusfearuth:
This pull request allows to use this library for Android >= 7.0, because of they use FileProvider for working with files.

PR #469, opened 2017-02-09
fix nougat

### LINKED ISSUE #376: [Android] Solution for FileUriExposedException in SDK 24
Starting with Android SDK 24 it is no longer allowed to expose the file: path. Instead it's recommended to use the content: URI. More info here:

https://developer.android.com/reference/android/os/FileUriExposedException.html

For now it's solved by not targeting SDK 24 but at some point this will not be an option anymore :)

skv-headless:
closes #376
http://stackoverflow.com/a/32710030/1410905 solution based on that answer.
@marcshilling I hope you can point cases which I need to test.

### DIFF
--- a/Example/android/app/build.gradle
+++ b/Example/android/app/build.gradle
@@ -77,13 +77,13 @@ def enableSeparateBuildPerCPUArchitecture = false
 def enableProguardInReleaseBuilds = false
 
 android {
-    compileSdkVersion 23
+    compileSdkVersion 25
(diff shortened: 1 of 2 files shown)


marcshilling:
What is the difference between this and what #464 is trying to fix?

rusfearuth:
First of all, It's OLD solution.
In my PR I use a new way.

marcshilling:
Closing in favor of #464

PR #474, opened 2017-02-17
fix(android): add option to wait for image/video to be stored before the callback

sondremare:
This functionality is present in the iOS part of the library, now it should work on all devices.


sondremare:
Had some issues on a LG phone, will fix

PR #479, opened 2017-02-22
Fix TypeScript module export

ligaz:
The current TypeScript declaration exports the module as a `default` export. This is incorrect because the module is exported as CommonJS single export and should be consumed with `export = require(...)` TypeScript's construct.

This change addresses the issue by exporting the module with `export =`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,6 @@
-
 declare module "react-native-image-picker" {
 
-    export interface Response {
+    interface Response {
         customButton: string;
         didCancel: boolean;
         error: string;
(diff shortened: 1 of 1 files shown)


ligaz:
Ping @hktonylee

hktonylee:
@ligaz Good job man! You reminded me how the import system works. Actually I am not familiar with it. Besides I am not working on react native now so it may takes a little longer for me to check.

wcandillon:
@hktonylee @ligaz This causes a regression since types likes `Response` cannot be imported. For instance it prevents to write the following:

```ts
ImagePicker.showImagePicker(options, (response: Response) => {
 // ...
});
```

Is there a way we could support both use cases? I'm actually not familiar with the issue this PR is fixing.

ligaz:
The issue that was fixed was that the typings were incorrect due to the way the module has been exported. The JavaScript for the module has it exported like CommonJS export: `module.exports = ...` whereas the TypeScript declaration was using a default export.

Maybe this could be fixed using namespaces?

ellman121:
@ligaz @JensDebergh Have you guys found a solution to this problem?  It is _very_ frustrating not being able to access types such as `Options` and `Response`

Issue #468, opened 2017-02-09
Duplicate image in folder Pictures


nhatchimai111:
![duplicate image](https://cloud.githubusercontent.com/assets/13510054/22770034/a87451d2-eebf-11e6-9709-4f64d6a7315b.png)

When I take photo then react-native-image-picker will create folder with 2 images same.

finch001:
this happen to me  my device is android7.1.1

rusfearuth:
@finch001 Could you check where first and second photos store? I guess why it could happen.

finch001:
it seem that Duplicate image all store in /storage/emulated/0/Android/data/com.photodemo/files/Pictures @rusfearuth

radjivF:
Any update on this one, I have the same problem in 2019 :'(

suvro24:
Any update on this one? I am still facing this issue.

username199327:
this also happened to the current app i'm working on, are there any updates on this?

DmitryGlobalbit:
Hi, I'm having the exact same issue. Any update regarding this bug?

imanshu1:
Still facing this issue in Motorola Phones

lucasR23:
i am also still facing this issue but in redmi note 7

AnurajDev:
This issue is still live!

Kamony:
had similar issue,
resolved by setting another storage destination path with option 

`storageOptions: {
    cameraRoll: true
  }`

devanshsadhotra25:
don't provide `storageOptions` and u won't see the issue

PR #491, opened 2017-03-01
Android: after resizing a newly captured image, delete the original image

egpast:
- Presumably, the caller doesn't want this image, and its path and URI are not returned in the response object, so its persistence seems unintentional.
- The current behavior is likely the cause of #468.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -438,6 +438,10 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
       }
       else
       {
+        // Newly captured images should only leave one copy on disk.
+        if (requestCode == REQUEST_LAUNCH_IMAGE_CAPTURE) { 
+          imageConfig.original.delete();
+        }
         uri = Uri.fromFile(imageConfig.resized);
         BitmapFactory.decodeFile(imageConfig.resized.getAbsolutePath(), options);
         responseHelper.putInt("width", options.outWidth);


marcshilling:
@rusfearuth can you check this out and merge if it looks good?

rusfearuth:
Okay. I'll check it.
вс, 5 марта 2017 г. в 3:53, Marc Shilling <notifications@github.com>:

rusfearuth:
@marcshilling Cloned that PR and found bug ) Probably, it's cool to remove useless photo, but when you chose some favorite photo from gallery, it'll be removed too )

I think it ain't good. Don't you?
I think that problem isn't there.

dsvictor94:
Maybe just check by the requestCode == REQUEST_LAUNCH_IMAGE_CAPTURE work arround this unexpected beravor

rusfearuth:
I think that System was auto scan media and caught original and resized images. I think we need to research in this way.

@dsvictor94 Yeah, It could be used as workaround.

egpast:
Sorry about that! I had only considered the direct capture use case. The commit now preserves images from the device library, using the `requestCode` check @dsvictor94 suggested.

rusfearuth:
@egpast I'm working on it. This commit is just workaround, problem is still not detected.

marcshilling:
@rusfearuth where does this stand now?

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -438,6 +438,10 @@ public void onActivityResult(Activity activity, int requestCode, int resultCode,
       }
       else
       {
+        // Newly captured images should only leave one copy on disk.
+        if (requestCode == REQUEST_LAUNCH_IMAGE_CAPTURE) {
```
rusfearuth: I agree with this Pull request, but I think that it should be located into MediaUtils as static method.
Please move this logic to `MediaUtils.removeOriginalIfNeeded(imageConfig, requestCode)` and use it at ImagePickerModule.

rusfearuth:
@marcshilling Added changes to the next my PR

Issue #485, opened 2017-02-24
latest version breaking on android with react-native 0.41.2


manjeets12:
I followed example and used following code, but it is just crashing  app as soon as I am clicking on use camera options on the other hand it is working perfectly when I am using select image from library.

> ImagePicker.showImagePicker(options, (response) => {
>   console.log('Response = ', response);
> 
>   if (response.didCancel) {
>     console.log('User cancelled image picker');
>   }
>   else if (response.error) {
>     console.log('ImagePicker Error: ', response.error);
>   }
>   else if (response.customButton) {
>     console.log('User tapped custom button: ', response.customButton);
>   }
>   else {
>     let source = { uri: response.uri };
>     this.setState({
>       avatarSource: source
>     });
>   }
> });

 "dependencies": {
    "react": "15.4.2",
    "react-native": "0.41.2",
    "react-native-image-picker": "^0.25.5"
  },

rusfearuth:
@manjeets12 Could you give more information about crash? I mean logs from Android.

Did you add the next lines to the `android/app/src/main/AndroidManifest.xml`
```xml
<uses-permission android:name="android.permission.CAMERA" />
<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```

manjeets12:
Yes I have added those lines, 
here is my code and errors image

> const imageOptions = {
>       quality: 1,
>       maxWidth: 500,
>       maxHeight: 500,
>       storageOptions: {
>         skipBackup: true
>       }
> };
> 
> ImagePicker.launchCamera(imageOptions, (response)  => {
> 		  // Same code as in above section!
> 		  let source = { uri: response.uri };
> 		    this.setState({
> 		      avatarSource: source
> 		    });
> 		});

![image](https://cloud.githubusercontent.com/assets/17762875/23302102/4f1f505a-fab3-11e6-87d9-30eb03ae255c.png)

rusfearuth:
@manjeets12 What version of Android did you use for testing? Was it device or emulator? If it was emulator, will you write emulator params for repeating bug?

manjeets12:
I am testing on my micromax mobile (android 5.1)

rusfearuth:
Thanks.

rusfearuth:
@manjeets12 Do you use proguard in your Android project?

manjeets12:
no I didn't , even I don't have a idea about that,

manjeets12:
[...]

Issue #478, opened 2017-02-20
Consistent permission denied behavior between iOS and android 


skv-headless:
**ios will call callback**
https://github.com/marcshilling/react-native-image-picker/blob/master/ios/ImagePickerManager.m#L203

**but android will not**
https://github.com/marcshilling/react-native-image-picker/blob/master/android/src/main/java/com/imagepicker/ImagePickerModule.java#L258

I can make PR but it may break someone's code.

rusfearuth:
I see. Will be fixed this week.

skv-headless:
It happens when select `never ask again`.

ptomasroos:
Looking forward to that @rusfearuth !

rusfearuth:
@ptomasroos I'm going to make PR in the next 2 days. Everything is ready, but I need to test my solution.

ptomasroos:
Great!

rusfearuth:
```
bug still on lastest version 0.25.7 (my device : android 4.4.2).
Crash on Camera option.
Debug log :
"02-27 18:18:19.858 29125-29172/com.package W/ContextImpl: Failed to ensure directory: /storage/extSdCard/Android/data/com.package/files/Pictures"
```

marcshilling:
@rusfearuth good to close this now?

rusfearuth:
I think yes.
вс, 5 марта 2017 г. в 3:55, Marc Shilling <notifications@github.com>:

PR #496, opened 2017-03-03
[ANDROID] Refactoring of Code plus bug fixes

rusfearuth:
[x] Fixed issue #485 (Updated README.md)
[x] Fixed issue #478 (Fixed permission flow)
[x] Refactored small part of code (not all)

I added explaining dialog for case when user checks `don't ask again`. That dialog allows to go app setting for switch on permissions.

Also I added support of customisation dialogs and updated view to modern.


dsvictor94:
It is a work in progress or is just waiting a review?
anyway, here my review:
In the README:
- I think now in Automatic Installation section the steps 2, 3 and 5 is required for Android
- Step 2 and 3 I think there can have a note to say its not required at reactnative 0.42

In android/src/main/AndroidManifest.xml
- I think merge this pull request will not revert the manifest

Another note is that when I upgraded to this version with `npm install git://github.com/marcshilling/react-native-image-picker.git#bugfix/478 --save` the error `ImagePickerManager.launchCamera got 2 arguments, expected 1` starts to throw, do you have changed the api interface?

rusfearuth:
@dsvictor94 sorry for inconvenient. I forgot that we could use this method without calling of `showImagePicker`. Now you can test it again.

PR #501, opened 2017-03-07
[UPDATE] Package name for manual installation

OzoTek:
I noticed the package and class name for the Android manual installation were wrong on the README and I managed to make it work by fetching this package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,16 +79,16 @@ IMPORTANT NOTE: You'll still need to perform step 4 for iOS and steps 2, 3, and 
 6. Add the import and link the package in `MainApplication.java`:
 
     ```java
-    import com.imagepicker.ImagePickerPackage; // <-- add this import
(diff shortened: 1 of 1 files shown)


marcshilling:
@OzoTek why do you say they are wrong? As far as I can tell, the package name is still `com.imagepicker.ImagePickerPackage`...

rusfearuth:
@OzoTek We haven't got `PickerPackage` and `com.reactnative.ivpusic.imagepicker` package name too. I see that you wrote to wrong project ;) You need to go [there](https://github.com/ivpusic/react-native-image-crop-picker). We're [react-native-image-picker](https://github.com/marcshilling/react-native-image-picker) ;)

OzoTek:
Wow, I guess I was tired sorry, will try to drink a coffee next time :p

PR #506, opened 2017-03-10
Android: fix cancellation handling

egpast:
- This was broken by #496.
- Currently, the response object is empty, lacking the `didCancel` flag.


marcshilling:
This is included in `v0.26.2` on npm, thanks!

antoinerousseau:
When I take a photo with my LG G3 (D855) that `resultCode` is `0` (`Activity.RESULT_OK` is `-1`) so `invokeCancel` is called when it should not

antoinerousseau:
I can't understand why, but if I remove the condition to let the code run anyway, I get a response with an image of width -1 and height -1, so I'm using this to detect a cancellation... @egpast @marcshilling any idea how to improve this?

PR #513, opened 2017-03-14
[ ANDROID ] Added support of storageOptions key

### LINKED ISSUE #508: Image doesn't appear in gallery after saving [Android]
When setting `storageOptions: {}` (such as in the example app), the docs say "the image will get saved in ... the Pictures directory on Android". I think it used to do this, but no longer.

Repro steps:

1. Run the example app on Android
2. Take a picture
3. Check your gallery app, it's not there
4. Check your Pictures folder, it's not there either

Tested on a Nexus 5X 6.0 emulator (using latest master 62cf9edbde45240a75b45c789042c9db488c6b76) and a Nexus 6P 6.0 physical device (using v0.25.7). I haven't tested on iOS because the simulator isn't able to take pictures.

---

Clarification: It sounds like currently the image is being saved under your app's `.../your_package_name/files/Pictures/`, whereas I assumed it would be saved under the shared `/sdcard/Pictures/`. Read below for more details about upcoming changes.

rusfearuth:
This pull request allows to fix #508 and contains refactoring.


marcshilling:
This is not how I want this to work. 

First of all, please don't change the Example project. There's no need to illustrate saving both locally and publicly.

On Android, we should leverage the existing `storageOptions.cameraRoll` option that is available on iOS, and if that is true the image should be saved to the public Gallery. Sorry for not chiming in about this sooner.

marcshilling:
@rusfearuth whats the status on this now that I've merged #516?

rusfearuth:
Still in processing. Refactoring of code will be pulled on this week.

orenklein:
@rusfearuth Anything new?

rusfearuth:
@orenklein still in progress

PR #516, opened 2017-03-16
Allow saving to camera roll on Android

cooperka:
Fixes https://github.com/marcshilling/react-native-image-picker/issues/508.

Note: Either https://github.com/marcshilling/react-native-image-picker/pull/513 or this PR should be merged. We needed this feature right away for our app, so I needed to do the work anyway and figured I'd submit a PR.

Tested and verified on Android emulator and physical device.

@marcshilling please let me know if you'd like anything tweaked. Note that the temp image is never deleted after resizing, but that's an existing bug that isn't changed by this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ noData | OK | OK | If true, disables the base64 `data` field from being generate
 storageOptions | OK | OK | If this key is provided, the image will be saved in your app's `Documents` directory on iOS, or your app's `Pictures` directory on Android (rather than a temporary directory)
 storageOptions.skipBackup | OK | - | If true, the photo will NOT be backed up to iCloud
 storageOptions.path | OK | - | If set, will save the image at `Documents/[path]/` rather than the root `Documents`
-storageOptions.cameraRoll | OK | - | If true, the cropped photo will be saved to the iOS Camera Roll.
+storageOptions.cameraRoll | OK | OK | If true, the cropped photo will be saved to the iOS Camera Roll or Android DCIM folder.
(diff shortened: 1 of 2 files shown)


cooperka:
@rusfearuth agreed, it could definitely be made more efficient. This works very well for my purposes though, and unfortunately I don't have enough time to improve it further. I hope it can be merged as-is and improved in the future. Thanks for the review.

rusfearuth:
@cooperka I like your solution. Only moving of file... But it's not big deal. Thanks for help!

@marcshilling I think we can merge this PR.

MilosR:
Is this tested? I got this on physical device: 
`ImagePicker Error:  Error moving image to camera roll: /app_images/Android/data/com.xxxxx/files/Pictures/image-953bd538-17e5-4976-8a0e-c6146c099799.jpg: open failed: ENOENT (No such file or directory)`

cooperka:
I'm using this successfully in production currently. Are you able to provide a demo app or specific instructions so we can try to reproduce the issue?

MilosR:
Everything is like in documentation, expect that i have used develop branch. When cameraRoll is set to false everything works perfectly.

```
import ImagePicker from 'react-native-image-picker';

var options = {
    title: 'Select image',
    noData: true,
    storageOptions: {
        skipBackup: true,
        cameraRoll: true,
    }
};

ImagePicker.showImagePicker( options, (response) => {
    console.log(response);   // response.error contains message 
});
```
Permissions in AndoidManifest are also set. Could cause be a custom rom installed on phone? (Android 6.0.1 - CM13). Unfortunately right now i do not have other device.

Looks like moveFile() is throwing an Exception or that other method below..

cooperka:
Hmm, I also happen to be using a physical Android 6.0.1 with CM13.0-20161220 (Nexus 6P) with no problem. Could you please open an issue for this so we can discuss there?

PR #535, opened 2017-03-31
Update buildToolsVersion  at build.gradle file

alexeypodolian:


rusfearuth:
I think it's rare case. Only if you add react-native to exist native project.

PR #541, opened 2017-04-04
Update realpath when file has been moved

MichelBahl:
if cameraRoll is set to true, an error occur on android due to moved image.

Fixed this issue by updating image path with path of the moved file.

### DIFF
--- a/Example/package.json
+++ b/Example/package.json
@@ -7,8 +7,8 @@
     "adb-reverse": "adb reverse tcp:8081 tcp:8081"
   },
   "dependencies": {
-    "react": "15.4.1",
-    "react-native": "^0.40.0",
+    "react": "15.3.1",
+    "react-native": "^0.33.1",
     "react-native-image-picker": "file:../"
   }
 }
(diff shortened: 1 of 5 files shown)


Example/package.json
```diff
@@ -7,8 +7,8 @@
     "adb-reverse": "adb reverse tcp:8081 tcp:8081"
   },
   "dependencies": {
-    "react": "15.4.1",
-    "react-native": "^0.40.0",
+    "react": "15.3.1",
+    "react-native": "^0.33.1",
```
marcshilling: This should not be changed. We made a decision a while back to require RN 0.40 or greater.

marcshilling:
Looks like we've got some conflicts based on recent changes.

@rusfearuth what are your thoughts on this?

alexustinovsm:
@marcshilling Going to check it out and write answer today.

rusfearuth:
@marcshilling Going to check it out and write answer today.

rusfearuth:
@marcshilling Will it work on iOS? I see @MichelBahl downgrade react-native version.

PR #548, opened 2017-04-10
Add fileSize support to video (Android only)

rayshih:


marcshilling:
This is not compiling when merged into `develop`. Please pull `develop`, rebase on top of that, and submit a new PR to `develop`

Issue #532, opened 2017-03-30
[Android] Image file size of 0 with Camera


SY102134:
Android4.4.2
picker0.26.2

I found that every time launch camera, will create a 0 size image in Android/Data/xxx/Pictures/image-xxx.jpg
When we finished taking pictures ,will write data in this 0 size image, if cancelled the process {didCancel=true}, but this 0 size image is not deleted

rusfearuth:
@wuhenwuya Does it happen when you cancel taken photo?

wuhenwuya:
yes, empty image file not delete

marcshilling:
Duplicate of #254

wuhenwuya:
@marcshilling i find that throws an exception no delete image in catch

rusfearuth:
@wuhenwuya It'll be fixed at next release. I've add fix for that bug in my local version of repository.

wuhenwuya:
@rusfearuth Thinks, If you could, i want to write the image url into the system library, when we open the image library by image picker, we can see the pictures we took before; but now i can not see!

rusfearuth:
@wuhenwuya Now, it's available from `develop` branch. I mean `writing the image url into the system library`

rusfearuth:
@wuhenwuya Could you check it out? We've rolled out 0.26.3 release.

wuhenwuya:
@rusfearuth very good，no problem in android4.4.2, image is also written to the system path; but have some minor problems:

1. because we first created 0 size image, if we finished taking pictures, but We do not click the Save button, but forced to close the app, or 0 size image still exist. so we can create image file after on press save button.
2. Image storage path in sdcard1/Pictures/image-xxx-xxx.jpg, not in app directory.

rusfearuth:
@wuhenwuya Okay. I'm going to rollout a new one release for 2 next days.

PR #554, opened 2017-04-13
Refactoring source code and adding old PRs

rusfearuth:
## Motivation

This pull request closes the next problem
* Removed of origin file after resizing #491 
* Changed algorithm of copy image files on `cameraRoll` option #491
* Fixed zero files on getting capture #532

## Test Plan

You need to take a photo. Then app asks you to confirm a taken photo, you'll need to move application to background and stop it.

After you need to run app again and choose select from gallery. The latest photo shouldn't be empty.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -32,6 +32,9 @@ android {
     lintOptions {
         abortOnError false
     }
+    testOptions {
+        unitTests.returnDefaultValues = true
+    }
 }
 
 repositories {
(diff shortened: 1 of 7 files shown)


rusfearuth:
@haitaoli how to contact with you about contribution model?

haitaoli:
@rusfearuth I created a gitter room to make all discussions visible to others: https://gitter.im/react-native-image-picker. You can log in with your github account.

haitaoli:
@rusfearuth what do you think if we keep the original and delete the resized one instead? The resized one may have much lower resolution and not very useful next time. If we support cropping (android-crop works fine with this module) in the future to match iOS functionality, there will be another intermediate file. The resized file cannot be deleted immediately. File.deleteOnExit() can do the job unless the app crashes. We can also consider using the same file name every time so we never leave more than one file behind.

PR #622, opened 2017-06-22
disable permission dialog when permissionDenied=null

skv-headless:
**motivation**

I don't want to use `permissionDenied` option because
- I can't customize dialog styles (we have custom)
- back button doesn't work on this dialog
- no ios support 

And now I can't disable it
 
**test plan**

add here https://github.com/react-community/react-native-image-picker/blob/develop/Example/App.js#L28
`permissionDenied: null`

previous behaviour - app crashes
current behaviour - nothing is happening


ptomasroos:
Can we get some attention here @marcshilling @janicduplessis

Its a very minor change

PR #633, opened 2017-07-06
permissionDenied for iOS

c19354837:

## Motivation (required)

iOS only. (Android has supported this festure, so this is iOS version. 

When the permission is denied at the first time, we cannot use this library until  we enable the permission in system settings. 

This PR provides a simple way to solve it by showing a alert view when the permission is denied, and user can click 'Open Setting'(default title) to open the app's settings directly.

All title in alert view can be customized. I have update the README.

You can disable this feature by make `permissionDenied` be null. (Only for iOS)

## Test Plan (required)

First, you should deny the permission at first time, also you can disable the permission in system settings if you have enabled it sometime.

Then click 'Take Photo…', the 'permission denied' alert view will show with two button('I'm sure' and 'Open Settings'). You can click 'Open Settings' if you want enable permission, or click 'I'm sure' nothing will happen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ IMPORTANT NOTE: You'll still need to perform step 4 for iOS and steps 2, 3, and 
     include ':react-native-image-picker'
     project(':react-native-image-picker').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-image-picker/android')
     ```
-    
+
 2. Update the android build tools version to `2.2.+` in `android/build.gradle`:
     ```gradle
     buildscript {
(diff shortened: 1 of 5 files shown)


adrianchinghc:
Would love to be able to implement something like this in my current app. How can I help to get this merged? @marcshilling @janicduplessis @c19354837

c19354837:
Just add this in your `package.json`, then run `npm install`

```
"react-native-image-picker": "git+https://github.com/react-community/react-native-image-picker.git#540eb07"
```

EricWiener:
This was really helpful. Thanks so much! To anyone else who needs this, after installing you can modify the alert by adding the following to the options object: 

```
permissionDenied: {
        title: 'Title of alert'
        text: 'Long message explaining what went wrong',
        settingTitle: 'Allow',
        okTitle: 'Later',
 },
```

vitorebatista:
Why has not this PR been approved yet?

squalvj:
is this PR approve yet ?

kyryloz:
@squalvj @vitorebatista 
why don't implemented it by yourself?
```javascript
if (response.error === 'Camera permissions not granted') {
  showNoPermissionDialog(
    'Your dialog message here',
  )
}
if (response.error === 'Photo library permissions not granted') {
  showNoPermissionDialog(
    'Or here',
  )
}
...

export const showNoPermissionDialog = title => {
  Alert.alert(title, null, [
    {
      text: 'Cancel',
      style: 'cancel',
    },
    {
      text: 'Settings',
      onPress: () => Linking.openURL('app-settings:'),
    },
  ])
}
```

chakrihacker:
I think @kyryloz solution works, rather than reimplementing alert and linking on native side, what are your thoughts @c19354837

PR #680, opened 2017-09-08
add 'showImageChooser' method (only for Android)

godness84:
## Motivation (required)

Currently if you want to let the user decide if to take a photo or to select from gallery, you should use the 'showImagePicker' method, that will show a dialog with the two options. After selecting the right option, the user chooses again which app to use. It's a 2 step process.

In order to make the process smoother and simpler, this PR adds a 'showImageChooser' method that proposes immediately to the user the apps that he can use in order to pick the image, either from camera or from gallery. Every app on the phone that can take a photo or grab it from gallery will be listed.

## Test Plan (required)

Tested on real device (Android 7.0)

screenshot_20170908-151451

If the user chooses the gallery app, the pic is taken correctly.
If the user chooses the camera app, the pic is taken correctly.
If the user chooses the document app and selects a wrong file (not an image), the promise is rejected with an error.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -3,11 +3,14 @@ package com.imagepicker;
 import android.Manifest;
 import android.app.Activity;
 import android.content.ActivityNotFoundException;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.pm.ResolveInfo;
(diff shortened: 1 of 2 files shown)


godness84:
If you don't like the idea to have a new method, it's to add an option to the 'showImagePicker' method.

godness84:
Ok, you didn't like the idea 😅

Johan-dutoit:
@godness84 😂  it was pre my time and with the decision to no longer support v2, I closed all open PR's (to make maintaining this repo easier).

PR #687, opened 2017-09-20
fix android saving to gallery AND temp & ios video more unique name

chan3111:
## Motivation (required)

For my company we are moving to react native and using this plugin to be able to capture images and videos. It is important that the image or video isn't saved to the gallery at all. The plugin works for that by not defining 'storageOptions' on IOS but for some devices on Android I found that it wouldn't always work. There is an issue with using putExtra on some devices. Essentially there are three different scenarios: 1. Camera only saves with the directory you define and pass through putExtra and doesn't save elsewhere. 2. Camera saves with the defined directory putExtra but also saves to the gallery at a seperate location. 3. Camera saves to ONLY the gallery location and ignores putExtra. I implemented a way to work around this issue and added an option for both video and picture 'noGallery' to ensure that if true the picture or video won't be saved to the gallery.

The second this I did was just finished the videoConfig (not necessary but nice to have) and sort of copied the way the images are taken for videos. So now the videos will return a fileName in the response object making it easier than having to use splice with the path to get it.

The third thing was just a simple change to the saved filename for IOS so that in a situation where you are uploading one video after the other and they are being displayed by filename, the name is unique and can be used as a key or atleast differentiated. It now uses the date and time with VIDEO_ prepended.

## Test Plan (required)

I tested this on three seperate Android devices, an old LG K4 (had saving gallery issue), a galaxy tablet and a Moto E (2nd Gen) (also had saving gallery issue) for the android changes.  I tested the IOS change on an IPad Mini (ios 9.3.x) and an IPhone 6 (ios 10.3.x). The only thing that I wasn't able to test was the third scenario of the putExtra situation. I only read that this issue happens but am not sure which devices do this, there is an implementation for it but I have no way of testing it. If anyone knows of such device or can test it for me, that would be greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ durationLimit | OK | OK | Max video recording time, in seconds
 rotation | - | OK | Photos only, 0 to 360 degrees of rotation
 allowsEditing | OK | - | bool - enables built in iOS functionality to resize the image after selection
 noData | OK | OK | If true, disables the base64 `data` field from being generated (greatly improves performance on large photos)
+noGallery | - | OK | If true, image or video will be stored only in the temporary directory on Android and will NOT be in your gallery
 storageOptions | OK | OK | If this key is provided, the image will be saved in your app's `Documents` directory on iOS, or your app's `Pictures` directory on Android (rather than a temporary directory)
 storageOptions.skipBackup | OK | - | If true, the photo will NOT be backed up to iCloud
 storageOptions.path | OK | - | If set, will save the image at `Documents/[path]/` rather than the root `Documents`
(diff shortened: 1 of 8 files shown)


dgurns:
Thanks for doing this! I need the same functionality. I tried your code and although images are now skipping the camera roll, they are still appearing in a "Pictures" directory as invalid files. 

Any idea why that's happening? 

I have "noGallery: true" and "storageOptions.cameraRoll: false"

chan3111:
It could be because you defined storageOptions. With the `noGallery: true`, you don't need to define anything for storageOptions.

dgurns:
Bingo. Thanks. Hope this gets merged soon!

hardcodet:
Any word on this? I think not having pictures you'll just need for temporary processing in the gallery (and then synced to DropBox etc.) is quite a substantial feature...

rollsroyc3:
Bump

PR #709, opened 2017-10-31
fix fileName undefined iOS 8

MaxToyberman:
## Motivation (required)

in iOS the app crashes https://github.com/react-community/react-native-image-picker/issues/640
## Test Plan (required)
open imagePicker , try to select an image (crashes on iOS 8).


gregogalante:
Get the same bug on IOS 11.2.5. Should this pull request be merged on a new version as soon as possible?

charpeni:
@gregogalante These changes seems to target iOS 8, I'm not sure how this will help on iOS 11.

janicduplessis:
@MaxToyberman React Native no longer supports iOS 8 so I'll close this. Feel free to use a fork if you really need to still support ios8.

PR #710, opened 2017-10-31
Set value for key "type" in picker response on iOS

sdduursma:
## Motivation

The response passed to the callback handler passed to `ImagePickerManager.showImagePicker` doesn't currently have a value for the key "type" on iOS. This is useful to have on iOS, just like it is useful on Android.

## Test Plan
Show the image picker like so:
```
ImagePickerManager.showImagePicker(options, (response) => { });
```
Just like on Android, `response` should have a value for the key "type" that is the MIME type of the picked media item ~~(although on iOS, this works for both images and videos)~~.


ios/ImagePickerManager.m
```diff
@@ -363,9 +363,13 @@ - (void)imagePickerController:(UIImagePickerController *)picker didFinishPicking
             NSData *data;
             if ([[[self.options objectForKey:@"imageFileType"] stringValue] isEqualToString:@"png"]) {
                 data = UIImagePNGRepresentation(image);
+                NSString *mimeType = (__bridge_transfer NSString *)(UTTypeCopyPreferredTagWithClass(kUTTypePNG, kUTTagClassMIMEType));
+                [self.response setObject:mimeType forKey:@"type"];
             }
             else {
                 data = UIImageJPEGRepresentation(image, [[self.options valueForKey:@"quality"] floatValue]);
+                NSString *mimeType = (__bridge_transfer NSString *)(UTTypeCopyPreferredTagWithClass(kUTTypeJPEG, kUTTagClassMIMEType));
+                [self.response setObject:mimeType forKey:@"type"];
```
haitaoli: These two blocks of code only differ in the enum value `kUTTypePNG` vs `kUTTypeJPEG`. Could you assign the value here then set response value out of the if/else blocks?

sdduursma: Thanks for the feedback. I pushed a commit that addresses this issue.

s123121:
This is incredible useful when use with upload via signed request to amazon s3, so have to know the type of image in order to upload. Please merge this!

PR #734, opened 2017-11-21
Add onPickerDidStart callback

foyarash:

## Motivation (required)

I want to be able, for example, to display an ActivityIndicator while i'm selecting an image or taking a picture. Naive method would be to set a loading state before calling showImagePicker, but this would show the indicator while the dialog is opened. Here it would display the indicator while chosing the image (but we wont see it), and while the response callback is being processed.

## Test Plan (required)

Create a simple view with an image and a button triggering the image picker. Apply the onPickerDidStart function which takes a callback as parameter, and set a loading state, which will show an ActivityIndicator. On the response callback, change the state so the Indicator is hidden.


arochedy:
Awesome feature to improve the ux of chossing image !

PR #742, opened 2017-11-26
refactor permission location

gengjiawen:
Motivation: Permission used by library should be declared in library manifest. It will be automatically merged to the final file.


haitaoli:
If the app only needs the photo picker without camera support, then it doesn't need to declare `CAMERA`, correct?

gengjiawen:
But in most cases, you will need both. And in android we use runtime permission.

PR #754, opened 2017-12-14
Fix the fact that the library asks for both permissions (camera and external storage) on Android

mreda7:
The library now in Android asks for both permissions ( camera and external storage ) in the same time when wanting to access gallery for example.
This behaviour is impacting the quality of the user experience .

## Motivation (required)
solving those issues : 
https://github.com/react-community/react-native-image-picker/issues/717
https://github.com/react-community/react-native-image-picker/issues/752
https://github.com/react-community/react-native-image-picker/issues/753

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -545,12 +545,39 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
     final int cameraPermission = ActivityCompat
             .checkSelfPermission(activity, Manifest.permission.CAMERA);
 
-    final boolean permissionsGrated = writePermission == PackageManager.PERMISSION_GRANTED &&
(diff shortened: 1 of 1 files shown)


winkelsdorf:
@charpeni That should only be necessary if the photo is actually stored outside one of the temp folders, isn't it?

chengsam:
@winkelsdorf If you provide the storageOptions for launchCamera.

chris-muller:
Any progress on this? Would be a nice improvement to my app if users weren't asked for camera permissions that I shouldn't need to ask for.

GoMino:
What is the status on this PR ?
This is a needed feature indeed

ashwinibm:
@Johan-dutoit Was this fixed recently?

Johan-dutoit:
@ashwinibm in v3 of the library, permissions requests been removed.  They're no longer required (see the readme).

Issue #783, opened 2018-01-23
Implicit boolean conversion of Objective-C object literal always evaluates to true


ayushiSharma22:
### Description

I upgraded my react-native version to 0.43.0. And ran my app in xcode 9.2 but i get this error: "Implicit boolean conversion of Objective-C object literal always evaluates to true" on line 685: -> statement: "return @NO;"

### How to repeat issue and example

Run Xcode 9.2 and received this warning "Implicit boolean conversion of Objective-C object literal always evaluates to true" in File ImagePickerManager.m

### Additional Information

* React Native version: [0.43.0]
* Platform: [iOS]
* Development Operating System: [MacOS]
* Dev tools: [Xcode]

ajfigueroa:
Were you running the example app or did you install this as a dependency in an existing application?

PR #817, opened 2018-03-25
[iOS] Fix Xcode value conversion warnings

ajfigueroa:

## Motivation (required)

This fixes warnings that appear when targeting 32-bit architectures (iPhone 4S for example), the warnings are as follows:

screen shot 2018-03-25 at 6 11 36 pm

1. `Implicit conversion loses integer precision: 'long long' to 'unsigned long'` (2 occurrences)
    - `malloc` expects an `unsigned long`. `NSUInteger` in 32-bit architectures is an `unsigned long` where in 64-bit it is a `unsigned int`, it makes more sense to cast to `NSUInteger` explicitly.
2. `Implicit pointer to integer conversion returning 'NSNumber *' from a function with result type 'BOOL' (aka 'signed char')` (1 occurrence)
    - This could introduce subtle bugs if any method depends on this result since we're casting an object to a BOOL. This will perform an existence check and `@NO` is an NSNumber object that is not nil so it'll always return `YES`.
    - This was also reported in #783 

### Scenario 1 (corresponds to point 1 above)

Pre-requisite: Find a GIF online and save it to Photos
1. Launch image picker via `Select a Photo` and tap `Choose from Library...`
2. Select the GIF that you downloaded earlier
3. Verify that it resizes correctly

### Scenario 2

This one is a bit more complicated to test for the following reasons:
- The returned `BOOL` from the `addSkipBackupAttributeToItemAtPath:` method is unused
- There is a direct coupling with the `[NSFileManager defaultManager]` singleton so we would need to refactor this and allow for dependency injection so we can mock the file manager and have a proper test for this

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -313,8 +313,9 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
                 ALAssetsLibrary* assetsLibrary = [[ALAssetsLibrary alloc] init];
                 [assetsLibrary assetForURL:imageURL resultBlock:^(ALAsset *asset) {
                     ALAssetRepresentation *rep = [asset defaultRepresentation];
-                    Byte *buffer = (Byte*)malloc(rep.size);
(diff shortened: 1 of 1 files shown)


ajfigueroa:
I've rebased my PR with latest develop

PR #819, opened 2018-03-26
[iOS] Update permissions flow for iOS 11 UIImagePickerController

### LINKED ISSUE #818: Add support for iOS 11 UIImagePickerController permissions change
### Description

In iOS 11, the `UIImagePickerController` is more sandboxed now meaning that asking for permission to access a photo is no longer required as it is a separate process.

To copy the excerpt from this SO [answer](https://stackoverflow.com/a/46594741/1631577) from rickster:
```
In iOS 11, UIImagePickerController runs as a separate process from your app. That means:

1. Your app can't see the user's whole Photos library — it gets read-only access just 
for whichever asset(s) the user chooses in the image picker.
2. Because of (1), your app doesn't need the standard privacy authorization for Photos 
library access. The user explicitly chooses a specific asset (or multiple) for use in your 
app, which means the user is granting your app permission to read the asset(s) in question.
```

For the below scenarios, build on an iOS 11 device.

### Scenario 1
1. Launch image picker via `Select a Photo` and tap `Choose from Library...`
2. Verify that the iOS native permission prompt appears asking for access to Photos
3. This shouldn't appear anymore

### Scenario 2
1. Launch image picker via `Select a Photo` and tap `Choose from Library...`
2. Verify that the iOS native permission prompt appears asking for access to Photos
3. Tap `Cancel`
4. Observe nothing happens
5. This shouldn't affect whether the image picker appears for iOS 11

### Solution

To address this issue, we just need to limit the permissions request to pre-iOS 11 devices for image picker single image selection only

### Additional Information

* React Native version: 0.40.0
* Platform: iOS
* Development Operating System: MacOS
* Dev tools:  Xcode 9.2, iOS 11 SDK

ajfigueroa:

## Motivation (required)

Fixes #818 which explains in more detail the rationale behind this change

#### Scenario 1
1. Launch image picker via `Select a Photo` and tap `Choose from Library...`
2. Verify the permissions prompt appears as normal
3. Ensure that tapping "Ok" lets you access Photos as per normal

#### Scenario 2
1. Launch image picker via `Select a Photo` and tap `Choose from Library...`
2. Verify the permissions  prompt appears as normal
3. Ensure that tapping "Cancel" does not let you access Photos

### On iOS 11 and up devices

1. Launch image picker via `Select a Photo` and tap `Choose from Library...`
2. Verify no permissions prompt appears
3. Ensure that selecting a Photo works as normal

### Xcode Versions

Ensure that the compiler macros work for both Xcode 9 (which has iOS 11 SDK) and all others below.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -205,6 +205,21 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
     else { // RNImagePickerTargetLibrarySingleImage
         // Wrapper for permission check and showing picker
         void (^showPickerViewControllerIfPermitted)() = ^() {
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
+            if (@available(iOS 11.0, *)) {
+                // iOS 11 no longer requires permission to access single image
+                showPickerViewController();
+            } else {
+                [self checkPhotosPermissions:^(BOOL granted) {
+                    if (!granted) {
+                        self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
+                        return;
+                    }
+                    
+                    showPickerViewController();
+                }];
+            }
+#else
             [self checkPhotosPermissions:^(BOOL granted) {
                 if (!granted) {
                     self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
@@ -213,17 +228,10 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
                 
                 showPickerViewController();
             }];
-        };
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
-        if (@available(iOS 11.0, *)) {
-            // iOS 11 no longer requires permission to access single image
-            showPickerViewController();
-        } else {
-            showPickerViewControllerIfPermitted();
-        }
-#else
-        showPickerViewControllerIfPermitted();
 #endif
+        };
+
+        showPickerViewControllerIfPermitted();
     }
 }
 
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -214,7 +214,7 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
                 showPickerViewController();
             }];
         };
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
         if (@available(iOS 11.0, *)) {
             // iOS 11 no longer requires permission to access single image
             showPickerViewController();
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -203,35 +203,19 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
         }];
     }
     else { // RNImagePickerTargetLibrarySingleImage
-        // Wrapper for permission check and showing picker
-        void (^showPickerViewControllerIfPermitted)() = ^() {
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
-            if (@available(iOS 11.0, *)) {
-                // iOS 11 no longer requires permission to access single image
-                showPickerViewController();
-            } else {
-                [self checkPhotosPermissions:^(BOOL granted) {
-                    if (!granted) {
-                        self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
-                        return;
-                    }
-                    
-                    showPickerViewController();
-                }];
-            }
-#else
+        if (@available(iOS 11.0, *)) {
+            // iOS 11 no longer requires permission to access single image
+            showPickerViewController();
+        } else {
             [self checkPhotosPermissions:^(BOOL granted) {
                 if (!granted) {
                     self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
                     return;
                 }
-                
+
                 showPickerViewController();
             }];
-#endif
         };
-
-        showPickerViewControllerIfPermitted();
     }
 }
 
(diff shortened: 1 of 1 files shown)


ios/ImagePickerManager.m
```diff
@@ -203,14 +203,27 @@ - (void)launchImagePicker:(RNImagePickerTarget)target
...
+            [self checkPhotosPermissions:^(BOOL granted) {
+                if (!granted) {
+                    self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
+                    return;
+                }
+                
+                showPickerViewController();
+            }];
+        };
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0
```
ajfigueroa: I included this macro check as I don't want to assume someone will only be using this on Xcode 9 and up.

ios/ImagePickerManager.m
```diff
@@ -203,14 +203,27 @@ - (void)launchImagePicker:(RNImagePickerTarget)target
...
+            [self checkPhotosPermissions:^(BOOL granted) {
+                if (!granted) {
+                    self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
+                    return;
+                }
+                
+                showPickerViewController();
+            }];
+        };
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
```
ajfigueroa: I included this macro check as I don't want to assume someone will only be using this on Xcode 9 and up.

getaaron: I think it would make more sense to include this logic within `showPickerViewControllerIfPermitted`

ajfigueroa: Were you envisioning something along this?
```objective-c
    // Wrapper for permission check and showing picker
    void (^showPickerViewControllerIfPermitted)() = ^() {
#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
        if (@available(iOS 11.0, *)) {
            // iOS 11 no longer requires permission to access single image
            showPickerViewController();
        } else {
            [self checkPhotosPermissions:^(BOOL granted) {
                if (!granted) {
                    self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
                    return;
                }
                
                showPickerViewController();
            }];
        }
#else
        [self checkPhotosPermissions:^(BOOL granted) {
            if (!granted) {
                self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
                return;
            }
            
            showPickerViewController();
        }];
#endif
    };

    showPickerViewControllerIfPermitted();
```

getaaron: Yes, it seems cleaner since all the necessary permission checks are encapsulated in `showPickerViewControllerIfPermitted`. Looks great 👍

ajfigueroa: Sure, I can dig it. I don't like the repetition but I feel like that is unavoidable. I'll try to update this then later tonight.

Off topic: Do you know if this repo is still being maintained?

ajfigueroa:
I've rebased my PR with latest develop

janicduplessis:
@ajfigueroa I reverted this since it still showed the permission prompt but when displaying the image instead. Might be related to the way RN loads local images but I'm not sure. You can easily repro in the example app in the repo.

ajfigueroa:
# TLDR

The permission prompt is being called due to this [line](https://github.com/react-native-community/react-native-image-picker/blob/master/ios/ImagePickerManager.m#L300):

```obj-c
PHAsset *pickedAsset = [PHAsset fetchAssetsWithALAssetURLs:@[imageURL] options:nil].lastObject;
```

This would have be handled differently in order for my PR to work. The typical approach for native apps is to pass the image from the image picker back to the view controller. Since we're limited by the React Native bridge, this is not possible and we need to work with PHAsset to form a file uri which requires `Read and Write` permissions.

This is my understanding of it but let me know what you think! I'm sorry this PR couldn't go through but it was a good learning experience regardless.

# Writeup

I did some digging and I realized since we can't pass the image over the bridge we instead pass over a file uri. I put a breakpoint in [App.js](https://github.com/react-native-community/react-native-image-picker/blob/master/example/App.js#L48) and observed that a file url is being sent in the form:
```
file:///Users/Anonymous/Library/Developer/CoreSimulator/
Devices/79182D1B-87DA-48DE-BB9C-3658E7D4B6B2/data/Containers/
Data/Application/1D636FEC-5855-4173-9D9E-02576BE5F880/Documents/
DF652477-31AE-47B5-A199-7D68208DFDC7.jpg
```

However, nothing seems particularly wrong about this file url.

If you deny the permission, the image still shows and when you accept the image it shows in Settings as `Read and Write`. This indicates to me that something else is requesting `Read and Write` access. This is unexpected since the image was already retrieved (via implied automatic read permission).

example-settings

I wanted to do some more digging and I realized that the permission prompt is happening before the file uri gets passed back over the bridge as evidenced per this photo:

screen shot 2019-01-06 at 11 36 34 pm

I was able to narrow it down to this section of code:
screen shot 2019-01-06 at 11 43 32 pm

Specifically, it seems like this line is the culprit:
https://github.com/react-native-community/react-native-image-picker/blob/master/ios/ImagePickerManager.m#L300

```obj-c
PHAsset *pickedAsset = [PHAsset fetchAssetsWithALAssetURLs:@[imageURL] options:nil].lastObject;
```

After looking the documentation, Apple states in the documentation for `PHAsset` [fetching assets](https://developer.apple.com/documentation/photokit/phasset/fetching_assets?language=objc):

> The first time you call one of the methods listed in Fetching Assets, Photos automatically prompts the user for authorization.

I believe this authorization prompt requests `Read and Write` access.

The methods listed in Fetching Assets are defined here: https://developer.apple.com/documentation/photokit/phasset?language=objc

screen shot 2019-01-06 at 11 51 21 pm

The one in the red rectangle is the method we're calling in `ImagePickerManager` which explains the Permissions prompt that is shown after the image is selected.
 
As for what this means for this PR, it means that this PR is not valid since the permissions prompt will need to be called when accessed with PHAsset due to the limitation of passing the image over the bridge.

janicduplessis:
@ajfigueroa Thanks for investigating this! I wonder if it would be possible to avoid using the PHAssets method, seems to be used only for getting fileName and geolocation info. I guess things like location is sensitive so iOS won't let us query it without permissions. What we could do is to pass an option to fetch only the info that we can without requesting permissions, I guess that for most of the cases we don't need all of this info and skipping the permissions prompt is nice.

QuentinWangWJJ:
Hi, is that mean we no longer need to add NSPhotoLibraryAddUsageDescription and NSPhotoLibraryUsageDescription in info.plist? Will this cause any problem when I will submit my app for AppStore review? Thanks!

Issue #354, opened 2016-10-14
Open Front Camera Directly in Android


sstur:
According to the readme this package doesn't support opening directly to the front-facing camera on Android.

<img width="569" alt="screenshot 2016-10-14 16 47 38" src="https://cloud.githubusercontent.com/assets/369384/19383354/182e34dc-922e-11e6-9f3b-c17ea9dbe996.png">

Any plans to support this in the future? Is it even possible on Android?

yfuks:
It's possible.
But we have no plan on doing it in **Android** since it's badly supported with native camera

KingAmo:
what a pity...really hope it can be suported...

HarshitMadhav:
Any update on this? I want to open the front camera directly in Android.

pawan8525:
use[](https://react-native-community.github.io/react-native-camera/docs/rncamera)

PR #835, opened 2018-04-18
Resolve memory leak in iOS image picker

SRandazzo:

## Motivation (required)

Currently, in a ReactNative app that is using a "Native Navigation" solution which uses multiple UIViewControllers, using react-native-image-picker will retain the presenting ViewController, causing a memory leak and undesirable behavior.

In our case, since the ViewController was retained, so were the components, which were dispatching off Redux actions in the background

Additionally, I have moved the UIAlertController code _into_ the main thread. UIKit is marked as not being thread-safe, and should only be used on the main thread. 

## Test Plan (required)

I was unable to find a test suite on the iOS target. 

A good manual QA test would involve the following:
- Present a UIViewController modally ("ViewController B")
- Set a breakpoint on the dealloc method of ViewController B
- Have ViewController B present the image picker options ( simply presenting the AlertController is enough to trigger the retain cycle)
- Dismiss ViewController B
- Observe if the dealloc method is called on ViewController B. If it is called, the viewcontroller has properly deallocated. If it has not deallocated, there is a memory leak still (you can observe this on master)

In a proper unit test, I would recommend using a weak ref to ViewController B, and verifying that it is nil after the image picker is dismissed

## Additional comments:

Thanks for this library. 
I noticed 2 extra files generated by XCode included in this PR. They are not part of the .gitignore. If you would like them removed please let me know

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -9,7 +9,6 @@
 
 @interface ImagePickerManager ()
 
-@property (nonatomic, strong) UIAlertController *alertController;
 @property (nonatomic, strong) UIImagePickerController *picker;
 @property (nonatomic, strong) RCTResponseSenderBlock callback;
 @property (nonatomic, strong) NSDictionary *defaultOptions;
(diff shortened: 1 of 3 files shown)


mvayngrib:
@SRandazzo could you point out which lines cause/fix the memory leak?

SRandazzo:
Hello, completely forgot I submitted this :-D
We have been using our own fork with this patch for quite some time and all has been smooth

I'll try to walk through the code path carefully, but let me know if I can be clearer:

The issue stems from here: 
`@property (nonatomic, strong) UIAlertController *alertController;`
https://github.com/react-community/react-native-image-picker/pull/835/files#diff-d8a8dfce7b30e35397fe68d17667d8c5L12

The `ImagePickerManager.m` is retaining a strong reference to the UIAlertController that it creates and presents. We pass this UIAlertController to the `root` VC, in this case `RCTPresentedViewController`, who assumes responsibility for it. 

Once the AlertController is dismissed via one of the actions, it disappears, and *should* deallocate by virtue of nothing holding on to it (eliminating all references it holds as well), however, the `ImagePickerManager.m` is still holding it via the `@property`, and so the `root` is also not released

mvayngrib:
@SRandazzo thanks for explaining that!

mvayngrib:
@SRandazzo this is kind of a long shot, but did you guys ever figure out how to make image picker open the front camera on android? (#354)

SRandazzo:
@mvayngrib sorry wish I knew but our RN app is iOS only at the moment so we haven't dug into Android at all. If i were more familiar with Android myself I'd take a look!

mvayngrib:
@SRandazzo haha, no worries, thanks anyway :)

PR #873, opened 2018-06-20
Fix crash android when choose image with big size

truongxuantran:


cpjolicoeur:
Could you please update your commit to only include things you actually changed?  It's nearly impossible to review in this state, as the vast majority of the diff appears to be only whitespace changes or moving lines around.

truongxuantran:
@cpjolicoeur  i have updated only file changed. Because i format code with Android Studio, so it make code change so much. :D

truongxuantran:
@cpjolicoeur I have updated my pullrequest. Could you please check it and merge to develop if it ok ? I think it will help other dev avoid crash app when use your lib on Android.Thank you !

cpjolicoeur:
@truongxuantran I am not a maintainer on this package.  Unfortunately, it appears to be pretty inactive as of late with a lot of open issues and PR's.  I'd be happy to contribute on it if the current owners want.  In the meantime, my company has forked and is maintaining our own fork of it to at least make some updates and additions.  You are more than welcome to file your PR's there if you would like: https://github.com/mojotech/react-native-image-picker

Just know that you'd have to use our fork in your `package.json` file and they wouldn't make it back here to the mainline version unless they take action on it here as well.

charpeni:
@cpjolicoeur Not anymore, it should be more active from now. Your contributions are welcome, you can send your PRs, and we can get this merged.

charpeni:
I can't reproduce this issue, even with 10000 x 10000 images and ~20mb.

What are your reproduction steps?

Also, I'm not sure about the solution. I find this a bit arbitrary to set the max size to 400 and then optimize the image by 50%. I'd prefer a wiser solution that guess a new size until it work.

PR #875, opened 2018-06-20
Feature: add option formatToMp4 to convert .mov to .mp4 in iOS

rizkisunaryo:
## Motivation (required)

When iOS records a video, the format is `.mov`.
When we upload this `.mov`, we can't play it in Chrome browser and some Android devices.
That's why, this `.mov` needs to be converted to `.mp4`

## Test Plan (required)

Tried recording, picking video from gallery, and uploading video from iOS. No crash happened. And the video could be played in Chrome.


rkdavidson:
FWIW, We manually patched our `react-native-image-picker` via `patch-package` to include this PR's changes, and have been using it in production for ~1-2 months now with no issues.

We use it to convert `mov` files to `mp4` before uploading, as part of our app's chat feature.
This way, iOS users can send videos to all Android users in a consistent way. Previously, Samsung devices couldn't play `mov` files.

Johan-dutoit:
Can the images be removed, seems like 2 were added for no reason?

DISKONEKTeD:
so is not being merged?

patissier-boulanger:
not merged yet??

nica0012:
Any news here?

PR #878, opened 2018-06-25
Updated Example to work with rn 0.55.4

manask88:
## Motivation (required)

The sample app is broken. I've updated it so it works with the latest stable rn version (0.55.4)


charpeni:
Fixed in 979b9a8.

PR #887, opened 2018-07-09
Fixed wrong image orientation when saved to camera roll

### LINKED ISSUE #470: CameraRoll pictures are rotated 90 degrees when saved
iOS

Front camera is only correct when snapped with the device in Right orientation.

Rear camera is only correct when snapped with the device in Left orientation.

paweljaneczek:
## Motivation

Fixes #470 

## Test Plan (required)

On iOS, Call `showImagePicker` with options: 
```
storageOptions: {
        skipBackup: true,
        cameraRoll: true
}
```
After you take photo, photo should be saved to camera roll with correct metadata so that it doesn't show rotated in it.


jpudysz:
+1

dylansmurray:
Looks like this is still broken when the "allowsEditing" flag is passed -- see issue opened here: 

https://github.com/react-community/react-native-image-picker/issues/955

PR #899, opened 2018-07-24
allows  for projects to override build script versions

comann:
## Motivation 

Currently, there is no way to specify the buildToolsVersion, compileSdkVersion, reactNativeVersion,  targetSdkVersion, and minimumSdkVersion. For projects that target api 27 or later, this produces warnings of multiple api's being used.  This PR utilizes the defaults already inside the `build.gradle` when not specified.

## Test Plan (required)

An example of overriding default version, to be placed inside the project's build.gradle
```
allProjects{
  project.ext {
     compileSdkVersion=27
     targetSdkVersion=27
     buildToolsVersion='27.0.3'
     minSdkVersion='21'
  }
}
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,6 +7,14 @@ def computeVersionName() {
     return json.version
 }
 
+def _ext = rootProject.ext
+
+def _reactNativeVersion = _ext.has('reactNative') ? _ext.reactNative : '+'
+def _compileSdkVersion = _ext.has('compileSdkVersion') ? _ext.compileSdkVersion : 25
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -45,7 +53,7 @@ repositories {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:+"  // From node_modules
+    compile "com.facebook.react:react-native:${_reactNativeVersion}"  // From node_modules
```
charpeni: React Native is not published to maven anymore, using `+` is the recommended method.

charpeni:
Thanks, it has been fixed by 8187eaf1ca308c0737ed13e96f98bd0033ca160e.

Issue #563, opened 2017-04-27
Permissions weren't granted on Android


enguerranws:
I'm testing an app which is using react-native-image-picker on my Android devices. It's working on iOs, not on Android.

Here's what I've done:
```
export function getImageFromLibrary () {

  let options = {
    quality: 1.0,
    maxWidth: 500,
    maxHeight: 500,
    storageOptions: {
      skipBackup: true
    }
  };

  return dispatch => {

    dispatch(requestImageFromLibrary())

    return ImagePicker.showImagePicker(options, (response) => {

      let responseObj = {};

      if (response.didCancel || response.error) {
          responseObj = {
            success: false,
            errorMessage: response.error,
            uri: response.uri
          };
          dispatch(receiveImageFromLibraryErr(responseObj));
      }

      else {

        responseObj = {
          success: true,
          uri: response.uri
        };
        dispatch(receiveImageFromLibrary(responseObj));
        // let source = { uri: 'data:image/jpeg;base64,' + response.data };


      }

    });
  }

}
```
```
<TouchableOpacity onPress={this.props.getImageFromLibrary}>
            <Text>Choose picture</Text>
</TouchableOpacity>
```
Here's what happen:
- I can tap the Choose picture
- it opens the "Select a photo" menu
- whether I tap Take photo or Choose from library, it logs an error "Permissions not granted" 

I uninstalled the app / reinstalled it, checked my application permissions in Android settings, but it never ask me for more permissions.

What do I miss? Should I use another module, like https://github.com/yonahforst/react-native-permissions?

kevinmcampos:
Any lucky on this?

Checkout your AndroidManifest.xml and see if it has these lines of code:

```
<uses-permission android:name="android.permission.CAMERA" />
<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```

This library already handle Android permissions, you don't need any other permission library for use it.

cshen4:
I'm having the same problem and my AndroidManifest.xml looks like this: 
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.test"
    android:versionCode="1"
    android:versionName="1.0">

    <uses-permission android:name="android.permission.INTERNET" />
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" /> 
    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
[...]

PR #971, opened 2018-11-06
Split permission requests for Camera & Gallery

lizraeli:


sryze:
Hi @lizraeli, why did you close this PR? Was there a reason? I too think that this package needs to request specific permissions based on whether requesting to open the photo library or camera. Personally I only use it to ask users to pick a photo from their library, not camera, so I think this patch is useful.

Tatandiaz13:
I did have this error #563 and this changes  work for me !

jacquesdev:
@lizraeli - Can we please get this issue re-opened and merged? The issue seems like a bug, to be honest, and this will fix it.

lizraeli:
@jacquesdev I've reopened the PR. I am happy to merge this once it gets reviewed.

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -547,10 +553,16 @@ private boolean permissionsCheck(@NonNull final Activity activity,
...
 
-    final boolean permissionsGrated = writePermission == PackageManager.PERMISSION_GRANTED &&
-            cameraPermission == PackageManager.PERMISSION_GRANTED;
+    boolean permissionsGranted = false;
 
-    if (!permissionsGrated)
+    switch (requestCode) {
+      case REQUEST_PERMISSIONS_FOR_LIBRARY:
+        permissionsGranted = writePermission == PackageManager.PERMISSION_GRANTED;
+      case REQUEST_PERMISSIONS_FOR_CAMERA:
```
Johan-dutoit: You're missing a break statement here, which causes the wrong 'permission' to be checked when selecting image from library

chengsam:
I wonder why a 5k starred library would have this bug (or feature)?

android/src/main/java/com/imagepicker/ImagePickerModule.java
lizraeli: @Johan-dutoit fixed!

genuinezack:
Agreed this would be a good feature to have cc @Johan-dutoit and @janicduplessis

kayzenkayzen:
Please merge!

KianooshSoleimani:
any solution? split permissions is need

PR #996, opened 2018-12-05
Fix cancelButtonTitle's Info in Reference.md 

gacha1987:
## Motivation (required)
Fixed cancelButtonTitle's Info
This is because cancelButtonTitle currently does not allow null and empty string

## Test Plan (required)
I don't think it requires tests.


janicduplessis:
Should we just fix it so it can accept null or an empty string instead?

PR #1000, opened 2018-12-10
Upgrade Android dependencies, added Google maven repository

Doko-Demo-Doa:

## Motivation (required)

This PR solves missing many Android libraries being removed from maven central (thanks to Google, since they moved their libs to their own Maven repo).

It also updates gradle version to 4.6 and support version to 28 (latest, stable).

## Test Plan (required)

In my project, there are two libraries that met mentioned issues (missing libs from maven). One being this, and one being `react-native-google-signin`. Both of them have the same solution, and this is it.


qlerebours:
Thanks for submitting :+1: 
I think it should be accepted as fast as possible, because other packages are depending on this PR

android/build.gradle
```diff
@@ -10,18 +10,19 @@ def computeVersionName() {
 buildscript {
     repositories {
         jcenter()
+        google()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.+'
+        classpath 'com.android.tools.build:gradle:3.1.4'
```
newyankeecodeshop: Should this file have the `buildscript` entry? I thought the application controls the build tools. I believe that since this is a library, it should not need to include the `repositories` or `buildscript`. Those entries are in the application's `/android/build.gradle` file and are applied to all child projects.

gponsinet:
#1002

backmeupplz:
Please, merge it

flat 550x550 075 f

charpeni:
Thanks for the laugh @backmeupplz.

I'm looking at this at the moment, I'm just not sure if we should bump Gradle directly to the latest without further tests. Thoughts?

I'm testing it at the moment; I'll keep you updated.

newyankeecodeshop:
@charpeni I would align your settings with those in RN 0.57 since that is the current stable release.
https://github.com/facebook/react-native/blob/0.57-stable/local-cli/templates/HelloWorld/android/build.gradle

backmeupplz:
@charpeni oh, please, definitely test it extensively :) thank you for the effort!

charpeni:
`react-native-image-picker@0.27.2`has been released, does it work?

android/build.gradle
Doko-Demo-Doa: I only keep it because it was originally in the code from master branch, as I think removing it might cause side-effects. It doesn't harm anyway.

Doko-Demo-Doa:
@charpeni I personally think it's safe. I have bumped all of my projects's gradle to 4.4 and 4.6 since April and no problem so far, as long as you include `google()` maven repo.

New project templates created by `react-native-cli` now use gradle 4.4 too.

android/build.gradle
```diff
@@ -11,18 +11,19 @@ buildscript {
     repositories {
         google()
         jcenter()
+        google()
```
angly-cat: Why specify `google()` twice?

Doko-Demo-Doa: Should be fixed now.

Doko-Demo-Doa:
@charpeni Sure, it's done.

elieteyssedou:
Thanks @janicduplessis for merging this. Is any release already planned ?

janicduplessis:
@elieteyssedou Released as 0.28.0

elieteyssedou:
Thanks a lot, I've upgraded and everything is working fine.

PR #1010, opened 2018-12-18
Replace deprecated 'compile' gradle configuration with 'implementation'

hyochan:
According to [ReactNative@0.57](https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#057) default gradle version is now `4.4`.

## Motivation (required)

Sometimes, android won't build because of this legacy.

## Test Plan (required)
Try to link this library and see if there is any warning in build in`react-native@57` project.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -51,15 +51,15 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


janicduplessis:
Thanks!

PR #1042, opened 2019-03-01
Export the interfaces in TypeScript

matt-oakes:
## Motivation
This allows users of the library to get access to the types that the library uses to annotate their code. For example, it allows this pattern:

```javascript
import ImagePicker, { ImagePickerOptions } from "react-native-image-picker";

const options: ImagePickerOptions = {
  noData: true
};
ImagePicker.openImagePicker(options);
```

I have also added the `ImagePicker` prefix to the interfaces to make it easier to distinguish them from other `Options` interfaces.

## Test Plan
This can be tested by using the library with TypeScript and using the new types. I have tested this manually on my machine by modifying the type definitions in the `node_modules/react-native-image-picker/index.d.ts` file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 declare module "react-native-image-picker" {
 
-    interface Response {
+    interface ImagePickerResponse {
         customButton: string;
         didCancel: boolean;
         error: string;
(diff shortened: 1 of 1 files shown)


janicduplessis:
Thanks! 👍

PR #1049, opened 2019-03-11
Fix video in image edge case

evanjmg:

## Motivation (required)

Some devices allow video anyways in photo only mode e.g Huawei Mate 20 Lite. This causes the app to crash on this line.
```
Fatal Exception: java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=13002, result=-1, data=Intent { dat=content://media/external/video/media/179 flg=0x1 }} to activity {com.velocity.v4b.blme.staging/com.velocityapp.consumer.MainActivity}: java.lang.NullPointerException: Attempt to read from field 'java.io.File com.imagepicker.media.ImageConfig.resized' on a null object reference
       at android.app.ActivityThread.deliverResults(ActivityThread.java:5042)
       at android.app.ActivityThread.handleSendResult(ActivityThread.java:5086)
       at android.app.ActivityThread.-wrap20(Unknown Source)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2071)
       at android.os.Handler.dispatchMessage(Handler.java:109)
       at android.os.Looper.loop(Looper.java:166)
       at android.app.ActivityThread.main(ActivityThread.java:7555)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:469)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:963)
```

Instead, it should error appropriately. 

## Test Plan (required)
Use an android device at allows video and photo when specifying only photos in the intent. There are no type tests to really implement here; perhaps there's a bigger issue with the intent being sent, but I'm not a 'pro' Android dev or anything.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -455,7 +455,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
       
       // some devices allow video anyways in photo only mode e.g Huawei Mate 20 Lite
       if (imageConfig == null) {
-        responseHelper.invokeError(callback, "Video is not supported in image mode");
+        responseHelper.putString("error", "Video is not supported in image mode");
         return;
       }
       
@@ -476,7 +476,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
       }
     }
 
-    if (imageConfig.saveToCameraRoll && requestCode == REQUEST_LAUNCH_IMAGE_CAPTURE)
+    if (imageConfig != null && imageConfig.saveToCameraRoll && requestCode == REQUEST_LAUNCH_IMAGE_CAPTURE)
     {
       final RolloutPhotoResult rolloutResult = rolloutPhotoFromCamera(imageConfig);
 


maiwenchang:
I met the same.

tourze:
same problem

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -452,6 +452,13 @@ public void onActivityResult(Activity activity, int requestCode, int resultCode,
     else
     {
       imageConfig = getResizedImage(reactContext, this.options, imageConfig, initialWidth, initialHeight, requestCode);
+      
+      // some devices allow video anyways in photo only mode e.g Huawei Mate 20 Lite
+      if (imageConfig == null) {
+        responseHelper.invokeError(callback, "Video is not supported in image mode");
```
Johan-dutoit: Is a user selecting a video the only reason it can be null?

evanjmg: I don't know :/  - seems so as it's not an image

Johan-dutoit: If you can investigate this more and see if there are any other reasons could be null and then update the error message to something more generic (if needed), then I can get this merged.

Relavant code: 
https://github.com/react-native-community/react-native-image-picker/blob/53316af68445de02a274fb53a7f2cdd0a59a431a/android/src/main/java/com/imagepicker/utils/MediaUtils.java#L81-L200

artyorsh: @Johan-dutoit this is still an issue in 2020. I guess keeping this opened for one year because of not enough generic error message is also not a correct way of resolving this, because of tons of affected production apps. Any chance to get this merged?

MorganTrudeau: Just tested this code. It was still crashing for me.

I changed line 458 to `responseHelper.putString("error", "Video is not supported in image mode");`

and before calling `imageConfig.saveToCameraRoll` I checked for `imageConfig != null`

All is working well with this.

evanjmg: @MorganTrudeau is this what you mean? Does it work for you? I haven't touched React native in a couple months nor do have the phone to reproduce this anymore :/

PR #1056, opened 2019-03-29
<Fix>fileName is null if photo from camera

RoJoHub:

## Motivation (required)

The `fileName` in response is nil  if photo from camera.
Fixed it.

## Test Plan (required)

take a photo by device and response have key of `fileName`


DavOnGit:
Pls consider merge this @janicduplessis @Johan-dutoit

PR #1063, opened 2019-04-09
Add saveToCache to storage options

steelbrain:
## Todo

- [ ] Tests

## Motivation

The main intention is to provide a way for consistent behaviour across android and iOS while storing images temporarily.

Current behavior for when `storageOptions` is not specified

Android: Saves to pictures directory
iOS: Saves to temporary directory

Current behavior for when `path` is specified in `storageOptions`

Android: Saves to pictures directory under a subdir
iOS: Saves to documents directory under a subdir

This PR provides a consistent behavior across OS while using cached directories, so apps can perform cleanup after themselves and OS can free it when feels the need.

Behavior with this PR when `saveToCache` is specified

Android: Saves to cache dir (under subdir if specified)
iOS: Saves to cache dir (under subdir if specified)

The main benefit consumers get out of it, is OS managed cache, no need to request permissions for external storage on Android while still having custom path and consistent cache directories across OSes.


steelbrain:
Dropping because I don't use this package anymore and have no relevant usecase. Whoever wants can pick up the changes under MIT license and continue the work

PR #1072, opened 2019-04-22
Update index.js.flow

miyabi:
Fix flow error on launchImageLibrary.

## Motivation (required)

It causes a flow error like below:

> Cannot call ImagePicker.launchImageLibrary because no more than 1 argument is expected by function type [1].
> Cannot call ImagePicker.launchImageLibrary with object literal bound to callback because a callable signature is missing
in object literal [1] but exists in function type [2].

## Test Plan (required)

```
import ImagePicker, { type Response } from 'react-native-image-picker';

function launchImageLibrary() {
  ImagePicker.launchImageLibrary({}, (response: Response) => {})
}
```

and then run flow.


ralli:
+1

PR #1083, opened 2019-05-20
improved build gradle script

OleksandrKucherenko:
- extraction of rootProject properties (according to latest codestyle of RN)
- support of symbolic links as a project directory (live re-mapping of directory structures)

## Motivation (required)

https://github.com/react-native-community/react-native-image-picker/issues/1081

## Test Plan (required)

- connect RN module to your application
- create inside android folder sub-folder `modules`
- create symbolic link to `modules/react-native-image-picker` -> `react-native-image-picker/android`
- add the dependency in an android app project `settings.gradle` as `include ':modules:react-native-image-picker'`
- try to compile the project.

Expected: compilation should pass.
Unexpected: compilation failed with JsonSlupper exception - `can not resolve package.json file path`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+[*]
+charset=utf-8
+end_of_line=lf
+insert_final_newline=false
+indent_style=space
+indent_size=2
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,9 +1,26 @@
 import groovy.json.JsonSlurper
+import java.nio.file.Files
 
+/** Get file of the project */
+static File fileOfProject(prj, filePath) {
+    final projectDir = prj.projectDir
+    final resolvedDir = Files.isSymbolicLink(projectDir.toPath())
+        ? Files.readSymbolicLink(projectDir.toPath())
+        : projectDir.toPath()
+
+    return new File("${resolvedDir}", filePath)
+}
+
+/** Extract property from rootProject if defined, otherwise use fallback value */
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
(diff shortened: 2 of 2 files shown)


OleksandrKucherenko:
@janicduplessis @charpeni ^

OleksandrKucherenko:
@janicduplessis - how much time it can take to merge small improvements? No push, but will be great to have some timeline

.editorconfig
```diff
@@ -0,0 +1,6 @@
+[*]
```
Johan-dutoit: This file should be excluded

android/build.gradle
```diff
@@ -51,7 +70,8 @@ repositories {
 }
 
 dependencies {
-    implementation "com.facebook.react:react-native:+"  // From node_modules
+    //noinspection GradleDynamicVersion
+    api "com.facebook.react:react-native:+"  // From node_modules
```
Johan-dutoit: api is deprecated

.editorconfig
OleksandrKucherenko: its a high-level configuration of tab/space formatting for many IDEs. for RN its quite common that devs use many IDEs: Intellij, vscode, atom, etc. .editorconfig is the only cross-ide thing that recognized by all of them.

android/build.gradle
OleksandrKucherenko: Consider you have app module which uses lib1 as a library and lib1 uses lib2 as a library. Something like this: app -> lib1 -> lib2.

Now when using api lib2 in lib1, then app can see lib2 codes when using: api lib1 or implementation lib1 in the app module.

BUT when using implementation lib2 in lib1, then app can not see the lib2 codes.

src: https://stackoverflow.com/questions/44413952/gradle-implementation-vs-api-configuration

PR #1090, opened 2019-06-05
Fixed orientation on save in library

bockc:
Pictures taken by the camera were incorrectly saved in the iOS device library, they were rotated of 90° for the portrait ones and 180° for the landscape ones.

Tested on iPhone 6S and iPhone XS Max and iOS 11 and 12, works perfectly fine.


BaderSerhan:
What's the status of this?

PR #1097, opened 2019-06-13
Add width, height, isVertical to video capture

### LINKED ISSUE #607: only partial response data when using launchCamera
I need to get the height and width or at least isVertical of the video.
It is only in the response when picking from the gallery. 
When using launchCamera, I only get the video path.

Is there a way to get this information when launching the camera directly?

iwikal:
resolves #607


BaderSerhan:
What'sthe status of this?

nica0012:
Any news on this?

iwikal:
No idea if it still works. I never ended up using this in production.

PR #1101, opened 2019-06-17
Change `compile` to `implementation`

nabati:
Documentation fix. 

`react-native link` does this already, but the steps in the manual installation steps still uses the old keyword.


Johan-dutoit:
Thanks!

PR #1113, opened 2019-07-01
Add ImagePickerPermissionOptions typings

Kaakans:
## Motivation

The ImagePickerOptions type was missing the permission options.

## Test Plan

N/A

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,7 +41,7 @@ declare module "react-native-image-picker" {
         allowsEditing?: boolean;
         noData?: boolean;
         storageOptions?: ImagePickerStorageOptions;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -41,6 +41,7 @@ declare module "react-native-image-picker" {
         allowsEditing?: boolean;
         noData?: boolean;
         storageOptions?: ImagePickerStorageOptions;
+        permissionOptions?: ImagePickerPermissionOptions;
```
Johan-dutoit: permissionDenied?

Kaakans: Yes of course. Fixed!

Issue #1017, opened 2019-01-10
Got response "didCancel", - expected result - get normal object with image uri, path, type, etc...


hadson19:
### Description
I've used  `showImagePicker` to load user photo from gallery or camera

```     
ImagePicker.showImagePicker({
  title,
  storageOptions: {
    skipBackup: true,
    path: "images",
  },
  quality: 0.8,
  maxWidth: 720,
  maxHeight: 1280,      
}, response => {
  ....
})
```
Actual Result: 
 ` {didCancel: true}`

Expected result:
 Get image information (uri, path, type, etc...)
  e.g.
```
{
    data:"..."
    fileName:"test.jpg"
    fileSize: 2501
    height:380
    isVertical:true
    originalRotation:0
    path:"..."
    type:"image/jpeg"
    uri:"..."
    width:380
}
```
### How to repeat issue and example

- Install any File storage with access to photo (e.g. Explorer File Manager '9xgeneration')
- try to load user picture using method `showImagePicker`

e.g. 
```
  ImagePicker.showImagePicker({
      title,
      storageOptions: {
        skipBackup: true,
        path: "images",
      },
      quality: 0.8,
      maxWidth: 720,
      maxHeight: 1280
  },
  response =>  console.log('response', response)
)
```
- before picking any picture the new windows should apper to select storage

![pickput image](https://user-images.githubusercontent.com/4102119/50986818-7fd77f00-1510-11e9-9a91-93091f37d8e4.png)

- click on default gallery icon to pick up the picture (click on "Just once")
- select available image

The result:  ` {didCancel: true}`

### Additional Information
* React Native version: 0.57.2
* Platform: emulator Android 8.0
* Development Operating System: Windows
* Dev tools: Android Studio 3.2.1

It's also reproduciung on android 7.x

P.S. If we have only Gallery (which is selecting by default) all works fine.

hackdie:
im having same issue, i found out that happens when i have more than 1 app to choose. (as i can see you have also two)

elkinjosetm:
Same here, have anyone found a solution for this issue?

Jacky9425:
Hi, the issue still exist when i have more than one option of app selection to trigger photo gallery. Unless one of the app is set to default then the image is uploaded successfully.
Any chance i can redirect the component to one gallery app at a time rather than allowing the app to detect multiple app selections?

PedramVeisi:
Having the same issue.

Jacky9425:
[...]

PR #1114, opened 2019-07-04
Fixed bug that can occur when user has multiple gallery apps

johanforslund:

## Motivation (required)

Targeting #1017. This bug occurs on Android when user has multiple gallery apps installed on the phone and tries to choose from library. It seems to only occur when launchMode is set to 'singleInstance' in Android Manifest.

This fix starts an Intent Chooser which solves this problem when using 'singleInstance'.

## Test Plan (required)

Tried running it on Android Pie physical phone and on emulator.


mariot:
Hi!

Any news on this?

Thanks

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -347,7 +347,7 @@ public void launchImageLibrary(final ReadableMap options, final Callback callbac
 
     try
     {
-      currentActivity.startActivityForResult(libraryIntent, requestCode);
+      currentActivity.startActivityForResult(Intent.createChooser(libraryIntent, "Choose gallery app"), requestCode);
```
Johan-dutoit: Change "Choose gallery app" to come from the options instead and then I can merge this.

johanforslund:
Options now have a new option called 'chooseWhichLibraryTitle'. If not set, it uses the default title from OS.

Johan-dutoit:
Thanks!  Now just needs to be documented (https://github.com/react-native-community/react-native-image-picker/blob/master/docs/Reference.md#options) and added to the types (https://github.com/react-native-community/react-native-image-picker/blob/master/src/internal/types.ts#L34).

johanforslund:
Fixed!

mariot:
Great news! Thanks a lot

PR #1118, opened 2019-07-07
Copy file instead of move

algera:
iOS 13 beta is not allowing the `moveItemAtURL` command to execute which may cause trimmed clips to fail based on permission.

Instead of moving the file, copy it.

## Motivation (required)
Attempting to access a trimmed video clip on device is failing based on permissions for iOS 13 beta.

Instead of affecting the source file with a `move`, instead `copy` it.

## Test Plan (required)

To test, select a video clip from the library that was been trimmed on an iPhone running iOS 13 beta or greater.

An example fetch in JS could be: 
```
    async showImagePicker () {
      ImagePicker.showImagePicker({
        mediaType: 'video',
        noData: true,
        permissionDenied: {
          text: 'To choose and change profile and survey imagery.'
        },
        title: 'Select media',
        durationLimit: 30,
        allowsEditing: true
      }, async (response) => {
        if (response.didCancel || response.error) {
          console.log('response error: ' + response.error)
          return
        }
        console.log('response.uri: ' + response.uri)
      })
    }
```

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -456,8 +456,14 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
 
                 if (videoURL) { // Protect against reported crash
                   NSError *error = nil;
-                  // iOS 13 b2 may not allow write access to tmp on a trimmed video clip
-                  [fileManager copyItemAtURL:videoURL toURL:videoDestinationURL error:&error];
+
+                  // If we have write access to the source file, move it. Otherwise use copy. 
+                  if ([fileManager isWritableFileAtPath:[videoURL path]]) {
+                    [fileManager moveItemAtURL:videoURL toURL:videoDestinationURL error:&error];
+                  } else {
+                    [fileManager copyItemAtURL:videoURL toURL:videoDestinationURL error:&error];
+                  }
+    
                   if (error) {
                       self.callback(@[@{@"error": error.localizedFailureReason}]);
                       return;
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -456,7 +456,8 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
 
                 if (videoURL) { // Protect against reported crash
                   NSError *error = nil;
-                  [fileManager moveItemAtURL:videoURL toURL:videoDestinationURL error:&error];
+                  // iOS 13 b2 may not allow write access to tmp on a trimmed video clip
+                  [fileManager copyItemAtURL:videoURL toURL:videoDestinationURL error:&error];
                   if (error) {
                       self.callback(@[@{@"error": error.localizedFailureReason}]);
                       return;


creativetim:
Not sure of the release process, but it would be pretty awesome if this fix made it to both `0.x.x` and `1.x.x`.

maxschmeling:
I have verified that this does fix the issue. My only concern is that I'm not sure of the implications of always copying instead of moving... Is the file that is being moved a temp (after compression?) file that will be cleaned up by the system so we don't have to worry about *moving* it?

I'm not an iOS internals expert... anybody know exactly what is going on and any possible gotchas?

maxschmeling:
@janicduplessis @Johan-dutoit is there anything else that needs done here? As far as I can tell, until this is merged, iOS 13 is completely unsupported... unless I'm missing something.

I'm happy to work on this if there is anything more that needs done.

algera:
Also able to adjust as necessary (maybe clarify the comment?).  My understanding about this file is that it is within a temporary storage folder which iOS will empty during regular garbage collection.

maxschmeling:
If it helps, I ran it through the debugger and `videoURL` for a video I recorded with the Camera app and then tried to open with the picker is:

```file:///private/var/mobile/Containers/Data/PluginKitPlugin/BABCCDC1-8CFB-4F73-8A54-FB22182BD3DE/tmp/trim.7F71AA0D-6549-4D10-A999-30B1E015A2AE.MOV```

**Comment by maxschmeling** [comment]
Would it be better to do something like this so that if the file is writable, we move it like it was before... and if it isn't writable, we don't?

```
if ([fileManager isWritableFileAtPath:[videoURL path]]) {
  [fileManager moveItemAtURL:videoURL toURL:videoDestinationURL error:&error];
} else {
  [fileManager copyItemAtURL:videoURL toURL:videoDestinationURL error:&error];
}
```

veermangat:
Same issue, could we please merge this PR in?

manuelrojas:
Any updates?

kodayashi:
@Johan-dutoit , @janicduplessis - Looks like another review is required? I see some others approved changes earlier but the PR cannot be merged. Would appreciate a quick look, we're not able to pick videos on ios 13 at the moment. Thanks!

Sherange:
@Johan-dutoit , @janicduplessis guys any update ?

kodayashi:
@Johan-dutoit , @janicduplessis guys any update ? Could we get someone else to help jump in to review this if you're busy?

kodayashi:
@charpeni - Hi Nicolas, looks like you're a maintainer as well? Is there anyone else that can help us out?

https://github.com/react-native-community/.github/blob/master/MAINTAINERS.md

jstettner:
Are there any updates? This is a very large blocker for ios 13.1.

kayzenkayzen:
Please Merge!

matteodanelli:
Any update on this? It seems critical for several people

creativetim:
In the interim you can create a patch pretty easily using [patch-package](https://github.com/ds300/patch-package).

rkdavidson:
Bumping thread, would love to get this fixed without needing to fork/patch manually.

giancarlo88:
We have also run into this problem. Any chance it can be prioritised? It's affecting more and more users who update to iOS 13, and we've had to resort to a forked repo.

Johan-dutoit:
I'll review once I get in front of my desk again.  I feel like there was a merge that contained a similar change.  I'll report back.

lafiosca:
Thank you for looking into it @Johan-dutoit

Johan-dutoit:
Released in v2.2.1

PR #1153, opened 2019-08-09
Fix android x

truongxuantran:


Johan-dutoit:
v1.0.1 already supports AndroidX and the code is already on the master branch.

PR #1159, opened 2019-08-14
iOS doesn't need permissions for picking a single image

tomasharkema:

## Motivation (required)
iOS doesn't need permissions for picking a single image. See more https://stackoverflow.com/a/46594741.

We currently have an issue with this library, that if we want to select a single image, we need to bother the user with it, which is unnecessary. 

## Test Plan (required)

Permissions are notoriously hard to test, so this is something that needs to be done manually.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -204,7 +204,18 @@ RCT_EXPORT_METHOD(showImagePicker:(NSDictionary *)options callback:(RCTResponseS
         }];
     }
     else { // RNImagePickerTargetLibrarySingleImage
-        showPickerViewController();
+        if ([[[UIDevice currentDevice] systemVersion] floatValue] < 11) {
+            [self checkPhotosPermissions:^(BOOL granted) {
+                if (!granted) {
+                    self.callback(@[@{@"error": @"Photo library permissions not granted"}]);
+                    return;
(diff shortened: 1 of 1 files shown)


janicduplessis:
Do we still need to ask permissions for ios < 11?

tomasharkema:
I guess so yes. I'll try to add a iOS version check.

tomasharkema:
@janicduplessis done!

Jazqa:
This does not work with the newest iOS/SDK version. If built with the newest iOS SDK, the system will force a permission prompt after choosing an image. While the image can be used after denying the permission on iOS 11 & 12, iOS 13 freezes after denying the permission.

VLNTNA:
Any idea or workaround how to solve the permission bug for the newest iOS version?

@Johan-dutoit @janicduplessis @tomasharkema

Jazqa:
Reverting this PR would be the logical fix, as older iOS versions also prompt permissions when built with newer SDK

AliUZAR:
If you will not revert this we need to use version 1.0.1 , My application not working because of permission issue.

ayciceksahin:
What about `ImagePicker.launchImageLibrary`  it is still asking permission after chose video from gallery. v2.3.1 ?

PR #1167, opened 2019-09-09
Fixed additional Permissions Issue

learnyst:
We need to take only the required permissions from the user. For Accessing photo library we need only the Storage permissions. But in current code we are requesting for camera permissions also. In current pull request we have modified code to request only storage permission for the photo library and removed camera permission

Verified the photo library access with only storage permissions in android and it is working fine

Yes

yes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -46,7 +46,7 @@ repositories {
 
 dependencies {
     api "com.facebook.react:react-native:+"  // From node_modules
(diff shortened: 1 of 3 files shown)


janicduplessis:
Thanks for sending the PR, why was a new dependency on `com.droidninja:filepicker` introduced? Is it possible to avoid adding this? It seems like a lot just to avoid asking for one permission.

PR #1191, opened 2019-10-04
Removed unnecessary permission request on Android

khuongTiet:

## Motivation (required)
Currently, on Android, the library requests permissions for both the external storage and the camera. 

This causes unexpected behavior as there could be times where a user would want to only allow permission for one and not the other.

Currently, the library is only saving to the external storage no matter what, so the `EXTERNAL_STORAGE` permission will still be needed, even if the user only wants to access the camera. To resolve this, I have updated the logic that writes to the external storage to only write there if the `storageOptions.path` property exists. This allows the library to save the capture imaged locally and avoids needing that extra permission unless explicitly asked for.

This PR updates the code base to match the expected behavior without using any additional external libraries

## Test Plan (required)

Ran on actual Android device (Samsung S9) and verified that the permission accept and declined behaviors are as expected:
Accept: Only ask for the one requested and opens the requested resource
Decline: Dismisses the permission alert

If you have added code that should be tested, add tests.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -565,15 +565,24 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
                                    @NonNull final Callback callback,
                                    @NonNull final int requestCode)
   {
-    final int writePermission = ActivityCompat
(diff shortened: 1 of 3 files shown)


kayzenkayzen:
I think this is really important feature. Please Merge!

PR #1197, opened 2019-10-10
BUGFIX Android crash fix when decoding unsupported files types

Frank1234:
## Motivation (required)

Fixes a crash: https://github.com/react-native-community/react-native-image-picker/issues/1196 When resizing is enabled, picking an image file type that is not supported on Android crashes the app. For example some files with the `image/tiff` or `image/heic` mime type.

## Solution (required)

`BitmapFactory.decodeFile` returns null If the specified file name is null, or cannot be decoded into a bitmap. In this case `MediaUtils.getResizedImage` returns null, though its annotation was `nonNull`. As a result the `ImagePickerModule` will crash.

1. I changed the annotation of `MediaUtils.getResizedImage` to nullable.
2. I changed the `ImagePickerModule` to be able to deal with a null response from `MediaUtils.getResizedImage`.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -350,7 +350,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
       libraryIntent = new Intent(Intent.ACTION_PICK,
       MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
 
-      if (pickBoth) 
+      if (pickBoth)
       {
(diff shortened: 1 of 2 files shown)


LouiseBC:
Could someone please take a look at this? I need this fix too ❤️

LouiseBC:
@Johan-dutoit or @janicduplessis ?

thejoaov (approved):

afilp:
I also need this fix, thanks a lot!

afilp:
Is this not needed anymore, this is why it is closed?

PR #1202, opened 2019-10-17
Fix Android Q file creation

amadeu01:
`getExternalStoragePublicDirectory` method is deprecated on Android Q, which means the folder is never created on Android Q devices, so we can not take a picture.

## Motivation (required)

Cannot take a picture with Android Q devices.

## Test Plan (required)

Run the demo app with Android Q device and use the following options on js code:

```
const options = {
      title: 'Select the new video',
      takePhotoButtonTitle: 'Record video...',
      storageOptions: {
        skipBackup: true,
        path: 'images',
      },
      mediaType: 'photo',
    };
```

### DIFF
--- a/android/src/main/java/com/imagepicker/utils/MediaUtils.java
+++ b/android/src/main/java/com/imagepicker/utils/MediaUtils.java
@@ -45,11 +45,23 @@ public class MediaUtils
                 .append(".jpg")
                 .toString();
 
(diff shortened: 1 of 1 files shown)


sahil290791:
@amadeu01 build import is missing.

Also as you are checking for Q in BUILD_VERSIONS, the code does not run as buildToolsVersion is not set as 29.0.0. I am not sure if changing the buildToolsVersion to 29 will break other react native packages.

Issue #1138, opened 2019-07-30
mediaType:'photo'  , Huawei mobile phone settings will appear video! Choosing a video will crash!


CJELLYS:
"react-native-image-picker": "0.26.10",
"react": "16.3.1",
"react-native": "0.55.4",

erez-guesty:
Same here.

Any work arround?

cleoTang:
+1 。
Same here.

Any work arround?

rock8808:
+1 
same here 
"react-native-image-picker":"1.0.1",
"react":"16.8.3"
 "react-native": "0.59.10",
phone :huawei honor v10 
andrord version:9
emui: 9.1.0

rahulcs1989:
Hi, I also encountered the same problem with Huawei phones. I have given mediaType for selecting photos only and it works pretty well in all devices except for Huawei. So if the user select any video instead of photo, the app will crash. So is there any solution for this?

cjinghong:
Same here, doing `ImagePicker.showImagePicker` with `mediaType: 'photo` as options will crash the app without throwing any errors at all.

erez-guesty:
still an issue

alexzzn:
find the ImagePickerModule.java file in Android projiect,and add a code`    
if (pickVideo)
    {
      requestCode = REQUEST_LAUNCH_VIDEO_LIBRARY;
      libraryIntent = new Intent(Intent.ACTION_PICK);
      // 351 line add,  filter image
      libraryIntent.setType("image/*");
    }`, i guess the imagePicker has default setType is "image/* video/*"

alexzzn:
find the ImagePickerModule.java file in Android projiect,
``` 
    if (pickVideo)
    {
      requestCode = REQUEST_LAUNCH_VIDEO_LIBRARY;
      libraryIntent = new Intent(Intent.ACTION_PICK);
      libraryIntent.setType("video/*");
    }
    else
    {
      requestCode = REQUEST_LAUNCH_IMAGE_LIBRARY;
      libraryIntent = new Intent(Intent.ACTION_PICK, MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
      libraryIntent.setType("image/*");
      if (pickBoth) 
      {
        libraryIntent.setType("image/* video/*");
      }
    }
```
i guess the imagePicker has default setType is "image/* video/*"

mingxin-yang:
issue

marlti7:
[...]

PR #1266, opened 2020-01-10
mediaType:'photo' , can Choosing a video will crash! (#1138)

alexzzn:


jehajj:
The solution provided by @alexzzn solves the issue!!

Issue #1100, opened 2019-06-17
Can't save photos to private folder


DevCeN:
# Bug

The `MediaUtils#createNewFile` method is always called with a `false` as the last parameter in `ImagePickerModule#launchCamera`, making the `forceLocal` flag that controls storing photos to private or external folder always goes to the external folder.

## Environment info

Library version: 0.28.1

## Steps To Reproduce

Describe what you expected to happen:

1. The `forceLocal` flag works as intended

PR #1282, opened 2020-02-03
Allow android to save photos to private directory

levipro:

## Motivation (required)

Added storageOptions.privateDirectory option so that we can save the photo in app's private files directory again on Android.

I have used this library since it was created, and once android support was added the inconsistency of the storageOptions between iOS and Android has been confusing.

In my app I want the photos to be saved to the app's private files directory. In older versions of the library I accomplished this by setting the storageOptions key in iOS and leaving it out on android. This provided the desired outcome of storing the photos in the apps private directory on both platforms, here is the old code the determined the file path (before version 0.27.0):

```
final File path = ReadableMapUtils.hasAndNotNullReadableMap(options, "storageOptions") && !forceLocal
                ? Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES)
                : reactContext.getExternalFilesDir(Environment.DIRECTORY_PICTURES);
```

Then in version 0.27.0 this code was changed and was undocumented, making it impossible for me to have android store the photo in the private files directory:

```
final File path = ReadableMapUtils.hasAndNotNullReadableMap(options, "storageOptions")
                && ReadableMapUtils.hasAndNotEmptyString(options.getMap("storageOptions"), "path")
                ? new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES), options.getMap("storageOptions").getString("path"))
                : (!forceLocal ? Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES)
                              : reactContext.getExternalFilesDir(Environment.DIRECTORY_PICTURES));
```

So this pull request was created to bring back the option of storing photos in the app's private file directory on Android.

The behaviour between iOS and Android is still not consistent but it is better documented now. At this point it could potentially cause lots of people issues by aligning the iOS and Android behaviour.

## Test Plan (required)

I have tested this on a Samsung Galaxy S8 and a Nokia phone. The behaviour is as described in the updated documentation. No change to iOS.


levipro:
Issue #1100

Johan-dutoit:
I'm not entirely against this change, we would just need to make it very clear in the release notes (would require a major version bump).

Let's leave it open a bit longer, to see if anyone has doubts about the change.  If there are no objections, then I'll merge and release.

Cheers.

levipro:
I just want to make it clear that this pull request will have no impact on people currently using it. The default behaviour is still the same as it has been since version 0.27.0. I have only added in an extra option for Android, and I have made the documentation more clear on the differences between iOS and Android

PR #1293, opened 2020-02-13
Fix Android URI permission denial bug when app refreshes

NdaJunior:
I propose that inside ImagePickerModule.java we use 

`libraryIntent = new Intent(Intent.ACTION_OPEN_DOCUMENT`

instead of 

`libraryIntent = new Intent(Intent.ACTION_PICK`

on line 356:

Screenshot 2020-02-13 at 14 55 35

for the following reasons:

As of Android M, certain permissions are granted temporarily to the activity that was running at the point when that permission was granted by the client. 

see: **https://stackoverflow.com/questions/37024668/issue-react-native-android-provider-permission-denial-exception**

see: **https://stackoverflow.com/questions/30572261/using-data-from-context-providers-or-requesting-google-photos-read-permission**

In the case of React Native Image Picker on Android, this is fine for once-off accessing and making use of an image that a client selects from Google Photos or any other photo library. I ran into an issue when a user accesses an image and I store the uri to that image on device. In this case, while the uri is saved on device and the user closes and reopens the app, an error will be throw on Android pertaining to denied permissions to the photos provider since the activity that originally requested the permission is no longer running.

Please see screenshot of the issue:

Screenshot 2020-02-13 at 12 02 54

By making use of "ACTION_OPEN_DOCUMENT", the uri permissions to the image is persisted across activities therefor the error is not raised

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -353,7 +353,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
     else
     {
       requestCode = REQUEST_LAUNCH_IMAGE_LIBRARY;
-      libraryIntent = new Intent(Intent.ACTION_PICK,
+      libraryIntent = new Intent(Intent.ACTION_OPEN_DOCUMENT,
       MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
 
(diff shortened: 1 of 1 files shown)


aplattinum:
I am having this issue as well. All of my photos are blocked from being opened from the device storage after I send the uri to firebase storage.  However, when using the camera, this isn't an issue.  Any updates on this??

PR #1417, opened 2020-09-09
Fix wrong package name in README

divykj:
## Motivation

Fixes the non-existent package name in `README.md` which is misleading and can sometimes be confusing.


janicduplessis (approved):
Thanks!

Issue #1393, opened 2020-07-13
Android 10: Camera Doesn't Open After Granting Permission


dmcrorieGIT:
# Bug report

## Summary
When using `ImagePicker.showImagePicker`, after granting permission for camera use selecting "Take a photo" does not open the camera.

[Related issue](https://github.com/react-native-community/react-native-image-picker/issues/1350#issuecomment-653143005)

<details>
<summary>On an Android 10 device (I used a Pixel 2 XL)</summary>

* Find a place in the code to kick off `ImagePicker.showImagePicker()`, a button might be easiest

* When the image picker kicks off, allow permissions for both camera and device storage

* Ensure that the camera does not launch, even though it should
</details>

## Environment info

`react-native info` output:

```bash
System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 166.34 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v10.16.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /Users/dustinmcrorie/.rvm/rubies/ruby-2.5.7/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  Languages:
    Java: 14.0.1 - /usr/bin/javac
    Python: 2.7.17 - /usr/local/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.11.0 => 16.11.0
    react-native: 0.62.0 => 0.62.0
  npmGlobalPackages:
    *react-native*: Not Found
```

Library version: 2.3.3

## Steps to reproduce

1. Use an Android 10 device, a Pixel 2 XL would be ideal if available
2. Find a place in the code to kick off `ImagePicker.showImagePicker()`, a button might be easiest
3. When the image picker kicks off, allow permissions for both camera and device storage
[...]

PR #1439, opened 2020-10-02
Fix android 10+ public folders access

iagormoraes:
## Motivation

There is a problem in this package where the picker and camera doesn't open as they can't have access to external directory folders, this PR uses scoped folders for android 10+.

## Test Plan

To test it, try to open the camera and shot a photo. For the image picker gallery, try to open it and select a image.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -416,7 +416,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule
         String realPath = getRealPathFromURI(uri);
         final boolean isUrl = !TextUtils.isEmpty(realPath) &&
                 Patterns.WEB_URL.matcher(realPath).matches();
-        if (realPath == null || isUrl)
+        if (realPath == null || isUrl || Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q)
         {
           try
           {
--- a/android/src/main/java/com/imagepicker/utils/MediaUtils.java
+++ b/android/src/main/java/com/imagepicker/utils/MediaUtils.java
@@ -11,6 +11,7 @@ import android.os.Environment;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import android.util.Log;
+import android.os.Build;
 
 import com.facebook.react.bridge.ReadableMap;
 import com.imagepicker.ImagePickerModule;
@@ -46,7 +47,13 @@ public class MediaUtils
                 .toString();
 
         // defaults to Public Pictures Directory
-        File path = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES);
+        File path;
+
+        if(Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+            path = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES);
+        } else {
+            path = reactContext.getExternalFilesDir(Environment.DIRECTORY_PICTURES);
+        }
 
         if (ReadableMapUtils.hasAndNotNullReadableMap(options, "storageOptions")) 
         {


mhashim6:
`externalDirectory` isn't discoverable by other apps, right? Users can't see these pics in their galleries for example/?

iagormoraes:
> `externalDirectory` isn't discoverable by other apps, right? Users can't see these pics in their galleries for example/?

As checked it is the path to `Android/data/data/your_package/`, differently from `getExternalStoragePublicDirectory` where it gives access to DCIM folder of a sd card for example.

mhashim6:
> > `externalDirectory` isn't discoverable by other apps, right? Users can't see these pics in their galleries for example/?
> 
> 
> 
> As checked it is the path to `Android/data/data/your_package/`, differently from `getExternalStoragePublicDirectory` where it gives access to DCIM folder of a sd card for example.

Yep that's what I meant. I think this is limiting, I assume many apps want their pics in the external public dir. So users can access them from other apps/galleries. 

Actually, if you want to achieve the same thing you can use the `privateDirectory` flag.

I think the only way to fix this while still allowing for external access is complying with scoped storage by using content providers.

iagormoraes:
> > > `externalDirectory` isn't discoverable by other apps, right? Users can't see these pics in their galleries for example/?
> > 
> > 
> > As checked it is the path to `Android/data/data/your_package/`, differently from `getExternalStoragePublicDirectory` where it gives access to DCIM folder of a sd card for example.
> 
> Yep that's what I meant. I think this is limiting, I assume many apps want their pics in the external public dir. So users can access them from other apps/galleries.
> 
> Actually, if you want to achieve the same thing you can use the `privateDirectory` flag.
> 
> I think the only way to fix this while still allowing for external access is complying with scoped storage by using content providers.

So currently we don't have access for DCIM on Android SDK higher than 28, making us having to make a scoped solution with temp folder, any photo taken by Camera API will still exposed outside, but unable to be read from inside.

Screen Shot 2020-10-02 at 15 24 33

Like @mhashim6 said, the proper solution would be to create providers for those versions.

gbferreira:
Foor me worked when i add this fix too: https://github.com/jkwiecien/EasyImage/pull/144/commits/d312d5e3a1bcd6f3363f0dd84c2777a6713ee3a6
using file provider. I used and device Samsumg S10 with android 10.


**ENVIRONMENT**
```
minSdkVersion = 17
compileSdkVersion = 29
targetSdkVersion = 29
```

jehajj:
+1

gnprice:
> I think this is limiting, I assume many apps want their pics in the external public dir. So users can access them from other apps/galleries.

I think many apps want the pics they create to be available from other apps. But the old way of doing that, by just placing them in the external public directory, has now been deprecated by Android and soon will not be available at all. (It's completely unavailable when targeting Android 11: https://github.com/react-native-image-picker/react-native-image-picker/issues/1393#issuecomment-661927130 ; and based on [the pattern of the last few years](https://support.google.com/googleplay/android-developer/answer/113469#targetsdk), Google Play will probably require all app updates to target Android 11 / API 30 starting about one year from now.) So the choices will be to make the images available in the new way, or not at all.

Here's [the documentation](https://developer.android.com/training/data-storage/shared/media) for the preferred way to do this. I believe the key steps are:
* Always create the file in a private location.
  * The WRITE_EXTERNAL_STORAGE permission no longer does us any good, and in fact we [should stop requesting it](https://developer.android.com/training/data-storage/shared/media#storage-permission) (using `android:maxSdkVersion` to conditionalize on the Android version.)
* When we want the file to be available to other apps, add it to the "media collection" called `MediaStore.Images` (or `MediaStore.Video` for videos.)
  * Once we do that, [other apps can see it](https://developer.android.com/training/data-storage/shared/media#scoped_storage_enabled) if they have READ_EXTERNAL_STORAGE permission.
  * The docs have [an example showing how to add to these collections](https://developer.android.com/training/data-storage/shared/media#add-item).

Effectively "the camera roll" no longer means a particular storage directory or directories. Instead, "the camera roll" is those two media collections, `MediaStore.Images` and `MediaStore.Video`.

I wrote a few more details about this at https://github.com/zulip/zulip-mobile/issues/4283 . In particular, if running into any trouble at the `readExifInterface` step, there's a note there about a way I think we can use `MediaStore` instead to get the same EXIF rotation data.

wodCZ:
Perhaps some of the responsibilities could be lifted to [library dedicated to cameraroll](https://github.com/react-native-cameraroll/react-native-cameraroll)?

I could imagine image-picker could only focus on picking file from library and capturing new photo/video to private storage.
For users that need the files to be available in the cameraroll (whatever that means in terms of 29+, in this case I refer to "Gallery" app), image-picker could mention rn-cameraroll's `save` method.

Just for clarification, as of now, they're having the same issue - android 10 is not fully supported.

gnprice:
@Johan-dutoit, could you say why you believe this is fixed?

I see there's [a beta of a new major version](https://github.com/react-native-image-picker/react-native-image-picker/releases/tag/v3.0.0-vnext.1). But I don't see anything in those release notes (or those of the later betas) about making the library compatible with scoped storage.

PR #1450, opened 2020-10-28
feat(typescript): add typings and more

Johan-dutoit:


Johan-dutoit:
@ravirajn22 Take a look when you can :)

README.md
```diff
@@ -34,29 +34,32 @@ Add permissions in `AndroidManifest.xml`:
 <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 ```
 
-
 # API Reference
 
 ## Methods
 
-### `launchCamera()`
+### `launchCamera()/launchCameraAsync()`
```
ravirajn22, line 41: For now lets stick with just `launchCamera()` where the response will come in callback. Two API's is maintenance headache. Users can easily wrap it up in Promise if they want.

README.md
```diff
@@ -34,29 +34,32 @@ Add permissions in `AndroidManifest.xml`:
...
+
+const response = await launchCameraAsync(options?);
 ```
 
 See [Options](#options) for further information on `options`.
 
 The `callback` will be called with a response object, refer to [The Response Object](#the-response-object).
 
-### `launchImageLibrary()`
+### `launchImageLibrary/launchImageLibraryAsync()`
```
ravirajn22, line 55: Same as `launchCamera()`.

src/index.ts
```diff
@@ -0,0 +1,57 @@
+import {NativeModules} from 'react-native';
```
ravirajn22, line 1: Instead of making the whole file typed which is unnecessary for simple javascript code and also creates trouble while development. Let's just export type definition for this library. Let all code be just plain javascript.

3GOMESz: > Instead of making the whole file typed which is unnecessary for simple javascript code and also creates trouble while development. Let's just export type definition for this library. Let all code be just plain javascript.
> 
> BTW I don't know typescript and expect lot of the users also not to. Why add extra dependencies when the simple way works.

TypeScript is great for any developer, any module update that changes the behavior of any method is automatically recognized by the user when looking at their IDE and already proposing a way to correct deprecated methods. It is simple to learn and at the same time it offers more code readability, better code auto completion, the possibility to better understand a code by looking at the limited types that any variable or function has. In my case I am passionate about TypeScript much more than pure JavaScript. In addition, any project written in TypeScript can be compiled in seconds for JavaScript, but in a JavaScript project, writing the types later for specific TypeScript users requires more work hours.

3GOMESz: But yes, lot of users still don't know ts

ravirajn22: We will export type declaration file that should do for people using typescript project.

Issue #1458, opened 2020-10-30
Remove buildToolsVersion from build.gradle or provide default


scousino:
### Description
Remove buildToolsVersion from the build.gradle. According to this [documentation](https://developer.android.com/studio/releases/gradle-plugin.html#3-1-0), after 3.1.0 the plugin automatically picks a version.

However, if its necessary to keep, then you should provide a default value as well as the extension. Here is an example of how you would do that:

![image](https://user-images.githubusercontent.com/8717527/97750358-ffa85200-1ac6-11eb-89ea-d4e69e6cb3fb.png)

This package will break builds on projects that do not specify a buildToolsVersion in their build.gradle ext section.

### How to repeat issue and example
* Install package in react native project
* Don't specify buildToolsVersion in the build.gradle ext section of that project
* Build fails

### Solution

Remove buildToolsVersion or specify a default fallback value.

### Additional Information

* Image Picker version: 3.0.0-vnext.2
* React Native version: 0.62.2
* Platform: Android
* Development Operating System: Windows

scousino:
It looks like in version 2.3.4, this default functionality was in place.

Johan-dutoit:
@scousino can you test directly from GH using the vnext branch?

scousino:
@Johan-dutoit tested and works

Johan-dutoit:
Thanks @scousino.

PR #1463, opened 2020-11-02
Remove unnecessary buildToolsVersion

ravirajn22:
Should fix https://github.com/react-native-image-picker/react-native-image-picker/issues/1458


Johan-dutoit:
Not worth a safeGet, as per GH-1458?

ravirajn22:
According to android docs `buildToolsVersion` no longer required so removed. Regarding the safeGet, all react-native project will have ext `compileSdkVersion` set in their build.gradle file. So felt unnecessary.

PR #1467, opened 2020-11-02
Fix: Android permission check

ravirajn22:
WRITE_EXTERNAL_STORAGE is not required for all cases. More info https://developer.android.com/training/data-storage

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,7 @@ If you are allowing user to capture image add `NSCameraUsageDescription` key als
 If you are allowing user to capture video add `NSCameraUsageDescription` add `NSMicrophoneUsageDescription` key also.
 
 ### Android
-
-Add permissions in `AndroidManifest.xml`:
-
-```xml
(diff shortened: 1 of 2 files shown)


Johan-dutoit:
Are you sure we don't need any additional permissions.  I'm busy updating the example to show permission requests and I needed to add permissions.

Johan-dutoit:
See:

image
image

Johan-dutoit:
Here's the [draft pr](https://github.com/react-native-image-picker/react-native-image-picker/pull/1469) for the above example app

ravirajn22:
Pull this change and try the example (stash all of your changes). It will work without any permissions.

Johan-dutoit:
To confirm, Android 9 and lower requires permissions, anything newer doesn't?

ravirajn22:
> To confirm, Android 9 and lower requires permissions, anything newer doesn't?

No it does not require. New changes you can find it here  https://developer.android.com/training/data-storage

Johan-dutoit:
LGTM.

ravirajn22:
Lets release a version

PR #1469, opened 2020-11-03
feat(example): update example styling

Johan-dutoit:


ravirajn22:
I left the example app to be simple purposely without even adding permissions and dependencies to easily update react-native versions and easy to understand. I highly recommend to avoid any custom components and dependencies.

BTW requesting permissions on iOS is handled by this library by default, only android permissions is to be handled by the user (PermissionsAndroid from react-native can be used for that purpose). After PR https://github.com/react-native-image-picker/react-native-image-picker/pull/1467 even android permissions is not required. Hence we can safely skip this PR.

Lets see if someone gets confused regarding these permissions and make changes accordingly to Readme.

Johan-dutoit:
Good point, I've removed it.  I'll leave keep the style changes though.

Issue #1466, opened 2020-11-02
Feature request for next: launchCamera for vertical images


jladuval:
I've been using the v1next project image picker and it's great! I would love to use launchCamera too but I need to take vertical images instead of horizontal ones. Is this possible already? If not it would be a great feature!

ravirajn22:
I don't understand by what you mean vertical and horizontal images. 

FYI, this library does not access camera API directly instead delegates it to the default Camera app. So you should probably be able to take any image.

jladuval:
Ok this could be my phone (Pixel 3). It seems like the orientation is off by 90 degrees. If I open the camera holding the phone vertically and take a photo vertically then the image comes out horizontal (width larger than length). If I flip my phone 90 degrees then take the image it comes out vertically (length larger than width). 

Perhaps this is related to #1338 and I should close this issue.

ravirajn22:
pixel android version

jladuval:
Android version 11 (Kernel version 4.9.223)

KeitelDOG:
Hi. How did you manage to get it to work with NextJS. I'm using React Native  with RN Web + NextJS, with next config. So far every complicated things have worked. But this package, with this import on Typescript `export * from './types';` can't work. I'm not using typescript yet. Do I need to install Typescript to get the Types?

Issue #1338, opened 2020-04-07
[Android] Camera image rotates 90 degress


krakzk:
I aml facing this issue, where my images taken from camera are rotated by -90 or 270 , which i cant confirm as my original rotation comes as 0 and isVertical is true.
i have tried the 360 rotation passing, giving quality 0.99 etc. Sadly none of them worked.
The issue doesn't appear in case of images selected from gallery.

Version : 0.28.1
Rn version : 0.59.3
Android device: Redmi note 8
Version : android 9

KrishyV:
I did manage to replicate it but only for selfies. The originalRotation from the response body is in fact 270 degrees. My proposed fix would be to import react-native-image-resizer and use the createResizedImage method to rotate the image by -90 degrees.

krakzk:
@KrishyV I had tried that, but it would affect the images selected from the gallery, as the issue is not there for them. Did u have any identification for the above select scenario that the image is from camera

KrishyV:
Yes, if you look at the URI from the Response body, it would differ if it is from Camera vs File System. You can do a `startsWith` on the URI to check if it is from taken from camera.

krakzk:
@KrishyV Thank you... Was able to sort out an algorithm based on that

krakzk:
If needed to resolve the issue without updating the version of your plugin, please change add the follwoing change in MediaUtils line 274
int orientation = exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_UNDEFINED);

This had worked in devices, but emulator still shows error in my case

PR #1472, opened 2020-11-03
Fix: image orientation issue on android

ravirajn22:
Fixes 
https://github.com/react-native-image-picker/react-native-image-picker/issues/1466
https://github.com/react-native-image-picker/react-native-image-picker/issues/1338
And should fix orientation related issues on android

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,4 +21,5 @@ android {
 dependencies {
     implementation "com.facebook.react:react-native:+"
     implementation "androidx.core:core:1.3.1"
+    implementation "androidx.exifinterface:exifinterface:1.3.1"
 }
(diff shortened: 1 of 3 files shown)


Johan-dutoit:
@jladuval please test these changes.

jladuval:
This works great @Johan-dutoit! Thanks

ravirajn22:
@Johan-dutoit we can merge it

ravirajn22:
Going forward every PR merge requires a release. If possible can you automate it @Johan-dutoit .

krakz999:
I'm on `3.0.1`, photos taken with `launchCamera` are still rotated by 90° when I post it to the backend. (other images are not)

ravirajn22:
First check if it's in proper orientation when used with Image component.

krakz999:
> First check if it's in proper orientation when used with Image component.

It is. Both gallery and photo images have the correct orientation in <Image/>. But only photos are rotated by 90° when uploaded.

ravirajn22:
So the problem is in your backend, it should consider the image orientation exif data from the image for processing

krakz999:
> So the problem is in your backend, it should consider the image orientation exif data from the image for processing. Google it.

Turns out `sharp` does not take the orientation tag into consideration without calling `rotate()` on it.
https://sharp.pixelplumbing.com/api-operation#rotate

Thanks for getting back!

Issue #1474, opened 2020-11-04
Failed on oppening photo camera (Android)


hiagoLF:
### Description

When I call ImagePicker.launchCamera() it does not open in photo option but in video option it woks as well.
.
### Additional Information

* React Native version: 0.63.3
* Platform: Android
* Development Operating System: Windows 10
* Dev tools: VScode

ravirajn22:
Require library version, mobile OS version, and your code to debug.

tamacroft:
me too, I think need to permission access camera. In version 2 is built in permission access pop up. But in version 3 did not show up pop up permission.

ravirajn22:
Version 2.x.x is not longer supported, please upgrade to 3.x.x
I hope the issue is solved, if not reply here or create new issue with details I am happy to fix it.

PR #1480, opened 2020-11-06
Fix: Android weird camera permission

ravirajn22:
Should fix https://github.com/react-native-image-picker/react-native-image-picker/issues/1474

Need to release a new version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ If you are allowing user to capture video add `NSCameraUsageDescription` add `NS
 
 No permissions required (`saveToPhotos` requires permission [check](#note-on-file-storage)).
 
+Note: This library does not require Manifest.permission.CAMERA, if your app declares as using this permission in manifest then you have to obtain the permission before using `launchCamera`.
+
 # API Reference
 
 ## Methods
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -30,6 +30,8 @@ If you are allowing user to capture video add `NSCameraUsageDescription` add `NS
 
 No permissions required (`saveToPhotos` requires permission [check](#note-on-file-storage)).
 
+Note: This library does not require Manifest.permission.CAMERA, if your app declares as using this permission in manifest then you have to obtain the permission before using `launchCamera`.
```
Johan-dutoit, line 33: Consider adding more emphasis about the whole library not requiring additional permissions.
As this comes up a lot in the issues.

android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -5,6 +5,7 @@
 import android.net.Uri;
 import android.os.Build;
 import android.provider.MediaStore;
+import android.util.Log;
```
Johan-dutoit, line 8: Guessing this isn't needed?

Johan-dutoit (changes requested):

Johan-dutoit:
Apart from the comments, looking good!

README.md
ravirajn22: Above this line I have mentioned `No permissions required`. So I am out of ideas.

ravirajn22: Lets see how the issues come then we will update appropriately. When in doubt better to avoid adding more text.

ravirajn22:
Can we merge and release a version?

Issue #1497, opened 2020-11-17
Cannot take video via launchCamera


scousino:
### Description
Cannot get the results from taking video via launchCamera. The camera app is launched and I can take video, but when I hit okay, instead of the callback being called I get a java.lang.RuntimeException saying "Failure delivering result"

![Screenshot_20201117-114546_ALIMqa](https://user-images.githubusercontent.com/8717527/99420243-fee72c80-28ca-11eb-9878-e7214f83d790.jpg)

### How to repeat issue and example
* Use launchCamera to take video. Record video and then hit ok

### Solution

Should be able to take video and receive the result on android

### Additional Information

* Image Picker version: 3.0.0-vnext.3
* React Native version: 0.62.2
* Platform: Android
* Development Operating System: Windows

scousino:
It looks like in onActivityResult in ImagePickerModule.java the data param is null, which is causing the null pointer exception that crashes our app.

scousino:
Device I'm using is running Android 9 if that helps

ravirajn22:
@scousino a PR is made to fix this issue. Try using it directly and check if the issue is fixed on your device. That would be helpful.

scousino:
I'm not sure how to test this directly, there is no branch for me to pull down with these changes

ravirajn22:
This is the branch https://github.com/ravirajn22/react-native-image-picker/tree/fix_videoCaptureUri

scousino:
Got it, thanks!

scousino:
@Johan-dutoit @ravirajn22 I saw the PR was merged in, but there isn't a new release version. When can I expect this?

ravirajn22:
point to master for now or that commit. @Johan-dutoit release ?

scousino:
Pointing to master is not a viable option for me. I need an actual npm package version release. Is this coming soon?

ravirajn22:
@Johan-dutoit can you make a release. 

Make this 3.x.x as the latest in npm so that new users install it. 

Remove the next string from the version names.

Can you add me to this library so that I can do release and minor commits.

scousino:
@Johan-dutoit anyone update on this??

Johan-dutoit:
@scousino try the latest 3.0.1.  Should be in there.

PR #1498, opened 2020-11-18
Fix:Uri null error in some cases when video captured from camera

ravirajn22:
data.getData() works in my android devices but may be not all android camera applications support it, hence use cameraCaptureURI like we do for image capture.

fixes https://github.com/react-native-image-picker/react-native-image-picker/issues/1497

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -167,9 +167,9 @@ public class ImagePickerModule extends ReactContextBaseJavaModule implements Act
 
             case REQUEST_LAUNCH_VIDEO_CAPTURE:
                 if (options.saveToPhotos) {
-                    saveToPublicDirectory(data.getData(), reactContext, "video");
+                    saveToPublicDirectory(cameraCaptureURI, reactContext, "video");
                 }
-                onVideoObtained(data.getData());
+                onVideoObtained(cameraCaptureURI);
                 break;
         }
     }


android/src/main/java/com/imagepicker/ImagePickerModule.java
```diff
@@ -167,9 +167,9 @@ public void onActivityResult(Activity activity, int requestCode, int resultCode,
 
             case REQUEST_LAUNCH_VIDEO_CAPTURE:
                 if (options.saveToPhotos) {
-                    saveToPublicDirectory(data.getData(), reactContext, "video");
+                    saveToPublicDirectory(cameraCaptureURI, reactContext, "video");
                 }
-                onVideoObtained(data.getData());
+                onVideoObtained(cameraCaptureURI);
```
scousino, line 172: This works, but would it be better to do a null check on data, and then use cameraCaptureURI only if its null?

ravirajn22: The fact is data.getData() and cameraCaptureURI should be same. So no need for checking.

ravirajn22:
@Johan-dutoit merge and release a new version

scousino:
To add some context, I looked a little more into this myself. I saw a few of StackOverflow posts mention that when you add an EXTRA_OUTPUT to an intent for the Camera, the Intent in onActivityResult will always be null. I couldn't find a reference to this in the docs other than this blurb about the camera intent: https://developer.android.com/reference/android/provider/MediaStore.html#ACTION_IMAGE_CAPTURE

Johan-dutoit (approved):

PR #1505, opened 2020-11-30
2.3.4

reberthkss:
## Motivation (required)

What was wrong:

Apps can't get permission to Write on external storage since Android API Level 29 due the new Scoped Storage introduced on Android 11 (https://developer.android.com/about/versions/11/privacy/storage), those restriction made cause the "open failed: EACCES (Permission denied)" error.

This workaround fixes the problem replacing the access to image file from the external/internal storage to app cache.

## Test Plan (required)

Simple image galery selection:

```
ImagePicker.launchImageLibrary({mediaType: "photo"}, async (response: any) => {
                const { uri, path, error } = response;

                if (!error) {
                   // success
                } else {
                    // error
                }
            });
```


ravirajn22:
May I know the reason for not upgrading to version 3.x.x? This android issue has been fixed in that.

reberthkss:
> 
> 
> May I know the reason for not upgrading to version 3.x.x? This android issue has been fixed in that.

i tested 3.0.0-vnext.3 and the same error occur. (EACESS: Permission Denied)

ravirajn22:
Sorry @reberthkss version 2.x.x are no longer maintained. You can fork and apply your changes and use it. Or you have to update to 3.x.x and create a new issue with details. I am happy to fix it.

Issue #713, opened 2017-11-04
App crash with error: Illegal callback invocation from native module. This callback type only permits a single invocation from native code.


pzawadzki82:
### Description
After invoking mutliple times ImagePicker.showImagePicker() my react-native based app crashed with the following error: Illegal callback invocation from native module. This callback type only permits a single invocation from native code. Stack tracke is attached below:

```
com.facebook.react.bridge.CallbackImpl.invoke CallbackImpl.java:30
com.imagepicker.ResponseHelper.invokeResponse ResponseHelper.java:76
com.imagepicker.ResponseHelper.invokeCustomButton ResponseHelper.java:56
com.imagepicker.ImagePickerModule.invokeCustomButton ImagePickerModule.java:497
com.imagepicker.ImagePickerModule$2.onCustomButton ImagePickerModule.java:193
com.imagepicker.utils.UI$1.onClick UI.java:66
android.support.v7.app.AlertController$AlertParams$3.onItemClick AlertController.java:1045
android.widget.AdapterView.performItemClick AdapterView.java:310
android.widget.AbsListView.performItemClick AbsListView.java:1145
android.widget.AbsListView$PerformClick.run AbsListView.java:306
android.widget.AbsListView$3.run AbsListView.java:3903
android.os.Handler.handleCallback Handler.java:739
android.os.Handler.dispatchMessage Handler.java:95
android.os.Looper.loop Looper.java:148
com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run MessageQueueThreadImpl.java:194
java.lang.Thread.run Thread.java:818
```

### How to repeat issue and example
Click many times on e.g buttom that invokes ImagePicker.showImagePicker()

### Solution
Currently no idea how ti fix it

### Additional Information

* React Native version: 0.48.3
* Platform: Android
* Development Operating System: Windows
* Dev tools: Android Studio and Atom

cma00000:
I have similar problem. Try to ask [PermissionsAndroid](https://facebook.github.io/react-native/docs/permissionsandroid.html#request) before imagePicker (Camera and Write External). Example:

```
import { PermissionsAndroid } from 'react-native';

async requestCameraPermission() {
  try {
    const granted = await PermissionsAndroid.request(
      PermissionsAndroid.PERMISSIONS.CAMERA,
      {
        'title': 'Cool Photo App Camera Permission',
        'message': 'Cool Photo App needs access to your camera ' +
[...]

PR #1510, opened 2020-12-04
Use callbacks collection to avoid multiple invocations on Android

agoldis:

- [ ] Match the **code formatting** of the rest of the codebase. (not sure, seems like it's formatted pretty well)

## Motivation

Wondering what do you think about this @ravirajn22.

The [`callback`](https://github.com/react-native-image-picker/react-native-image-picker/blob/main/android/src/main/java/com/imagepicker/ImagePickerModule.java#L30) is shared. When calling `launchImageLibrary` or `launchCamera` frequently, two Activities are getting started and the later invocation overrides `callback` for the former.

As a result, `callback` might be invoked more  than once, triggering 

```
java.lang.RuntimeException: Illegal callback invocation from native module. This callback type only permits a single invocation from native code.
...
```

Resolves https://github.com/react-native-image-picker/react-native-image-picker/issues/713

Re:  Why one would invoke image picker twice?

Shouldn't happen, but does happen accidentally due to various reason - e.g. RN badly handling guestures, slow devices, badly implemented React Hooks. 

The idea is that the library should be resilient to such kind of mistakes.

## Test Plan (required)

Run `launchCamera` or `launchImageLibrary` one after another, pick / create an image and see the code.

### DIFF
--- a/android/src/main/java/com/imagepicker/CallbacksCollection.java
+++ b/android/src/main/java/com/imagepicker/CallbacksCollection.java
@@ -1,0 +1,49 @@
+package com.imagepicker;
+
+
+import androidx.annotation.NonNull;
+import androidx.core.util.Pair;
+
+import com.facebook.react.bridge.Callback;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+
+public class CallbacksCollection {
+    final static  Map<Integer, Pair<Callback, Integer>> callbackMap = new HashMap<>();
+    final static Random random = new Random();
+
+    public static Integer set(@NonNull final Callback cb) {
+        int id = random.nextInt(65535);
+        callbackMap.put(id, Pair.create(cb, 0));
+        return id;
+    }
+
+    public static void invoke(final Integer cbId, Object... args) {
+        if (!callbackMap.containsKey(cbId)) {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


ravirajn22:
Thanks for the PR. The working of this library is undefined when you call it again before the previous call has finished.

The simple solution will be to set a boolean variable as true (for example `isPickerOpened = true`) before calling launchCamera or launchImageLibrary and set it `isPickerOpened = false` on callback called. In between if launchCamera or launchImageLibrary is called check if `isPickerOpened = true` then just return and proceed only when `isPickerOpened=false`.

I am not going to add this PR, may be I will see if I can add this in JS side of the library.

agoldis:
@ravirajn22 would be very kind of you

mohit23x:
> Thanks for the PR. The working of this library is undefined when you call it again before the previous call has finished.
> 
> The simple solution will be to set a boolean variable as true (for example `isPickerOpened = true`) before calling launchCamera or launchImageLibrary and set it `isPickerOpened = false` on callback called. In between if launchCamera or launchImageLibrary is called check if `isPickerOpened = true` then just return and proceed only when `isPickerOpened=false`.
> 
> I am not going to add this PR, may be I will see if I can add this in JS side of the library.

@ravirajn22 is this logic added..i couldn't find any reference of this in code.
The multiple invocations is leading to many app crashes.

PR #1533, opened 2020-12-13
Change directory where temporary files are stored to cache directory

ravirajn22:
Now we don't delete files on every call, instead it will be handled by
Android OS. So people can expect to save the Uri and use it in later
session unless android clears the cache.

Make a copy of image file picked from gallery into internal storage for
later access. We don't copy videos picked from android gallery, so the uri 
wont persist sessions.

Now the implementation is same in iOS and Android, except for handling
of videos in android where videos picked from gallery are not duplicated. Which needs further discussion.


ravirajn22:
@Johan-dutoit  commit and release.

Johan-dutoit (approved):

Issue #1546, opened 2020-12-17
videoQuality is not working.[🐛] 


SohaibKtb:
### Description
videoQuality is not working.

the same for 2 options on android either 'low' or 'high' videoQuality

the resolution still 144 pixels

```
let options = {
          videoQuality: 'high',
          mediaType: 'video',
          durationLimit: 30
        };
```

- Image Picker version: [``` "react-native-image-picker": "^3.0.0" ```]
- React Native version: [```"react-native": "0.63.3"```]
- Platform: [Android]
- Development Operating System: [Linux]
- Dev tools: [VSCode]

Johan-dutoit:
Hey, can you give this a test: https://github.com/react-native-image-picker/react-native-image-picker/pull/1549

SohaibKtb:
Hello @Johan-dutoit, I did a test and the same result persist,

this is the `Options.java` file in the `react-native-image-picker` package,

```
package com.imagepicker;

import com.facebook.react.bridge.ReadableMap;
import android.text.TextUtils;

public class Options {
    Boolean pickVideo = false;
    Boolean includeBase64;
    int videoQuality = 1;
    int quality;
    int maxWidth;
    int maxHeight;
    Boolean saveToPhotos;


    Options(ReadableMap options) {
        if (options.getString("mediaType").equals("video")) {
            pickVideo = true;
        }
        includeBase64 = options.getBoolean("includeBase64");
        
        String videoQualityString = options.getString("videoQuality");
        if(!TextUtils.isEmpty(videoQualityString) && videoQualityString.toLowerCase() != "high") {
            videoQuality = 0;
        }

        quality = (int) (options.getDouble("quality") * 100);
        maxHeight = options.getInt("maxHeight");
        maxWidth = options.getInt("maxWidth");
        saveToPhotos = options.getBoolean("saveToPhotos");
    }
}
```

and the options in my component still the same: 
```
let options = {
          videoQuality: 'high',
          mediaType: 'video',
        };
        ```

SohaibKtb:
@Johan-dutoit when I changed ```cameraIntent.putExtra(MediaStore.EXTRA_VIDEO_QUALITY, 1);``` in ImagePickerModule.java it works fine with a resolution `(18.7:9) 1080p`,

so I think there is a problem with getting the options value to ```cameraIntent.putExtra(MediaStore.EXTRA_VIDEO_QUALITY, "here" )```

what do you think?

[...]

PR #1549, opened 2020-12-21
fix(video-quality): 🐛 map video quality to relevant int

Johan-dutoit:
This defaults the video quality to high (1) and if the value is anything other than high (barring empty), then it will fallback to low (0).

Closes GH-1546

### DIFF
--- a/android/src/main/java/com/imagepicker/Options.java
+++ b/android/src/main/java/com/imagepicker/Options.java
@@ -1,11 +1,12 @@
 package com.imagepicker;
 
 import com.facebook.react.bridge.ReadableMap;
+import android.text.TextUtils;
 
 public class Options {
     Boolean pickVideo = false;
     Boolean includeBase64;
-    String videoQuality;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/imagepicker/Options.java
```diff
@@ -18,7 +19,12 @@
             pickVideo = true;
         }
         includeBase64 = options.getBoolean("includeBase64");
-        videoQuality = options.getString("videoQuality");
+        
+        String videoQualityString = options.getString("videoQuality");
+        if(!TextUtils.isEmpty(videoQualityString) && videoQualityString.toLowerCase() != "high") {
```
SohaibKtb: change
`videoQualityString.toLowerCase() != "high"`
to
`videoQualityString.toLowerCase().equals("low")` or `!videoQualityString.toLowerCase().equals("high")`

SohaibKtb (commented):
working now :+1:

Issue #1558, opened 2020-12-25
[iOS] Build failed with error: Undefined symbols for architecture x86_64 [🐛] 


papts0909:
### Description

I tried to upgrade & build my project on Xcode with version react-native-image-picker@3.1.2 and encountered error : 
`Undefined symbols for architecture x86_64:
"_OBJC_CLASS_$_ImagePickerUtils", referenced from:
objc-class-ref in libRNImagePicker.a(ImagePickerManager.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)`

So i unlink and link a package again with no luck

### How to repeat issue and example

- npm install react-native-image-picker@3.1.2
- react-native link react-native-image-picker
- Run build on Xcode => Encounter error

### Solution

I found 'ImagePickerUtils.m' is missing in 'RNImagePicker'.
After add 'ImagePickerUtils.m' to 'RNImagePicker', it worked again.
![image](https://user-images.githubusercontent.com/20726300/103125834-f69ac200-46be-11eb-84f8-ac8d233427cd.png)

At this time my project can run & build normally, but i dont know that's correct, so please help me check. Thanks you!

### Additional Information

- Image Picker version: 3.1.2
- React Native version: 0.59.10
- Platform: iOS
- Development Operating System: MacOS
- Dev tools: Xcode 11.6

tienle46:
I also has this bug in RN version 0.59.10. Pretty sure this will happen if we link the library manually

ravirajn22:
Sorry guys don't have time to support older versions, request you to fork this library and make these small changes and use it.

PR #1571, opened 2021-01-08
fix: [iOS] Build failed with error: Undefined symbols for architectur…

tienle46:
…e x86_64 (#1558)

## Motivation (required)

This pull request solved the problem: Build failed when link library manually on iOS with RN <0.60. This bug cause by missing ImagePickerUtils.m in Compile Source.

## Test Plan (required)

I have test this carefully on my project and see no problem with it.


ravirajn22:
hi, thanks for the PR. But we cannot test this library on older versions of react-native so I request you to make this change in forked version of this library and use it. Thanks.

PR #1597, opened 2021-01-26
[iOS] get access to the original photo path and name

GaspardC:

## Motivation 

On iOS photos in the picker are limited to 2048x2048px
You may need the photo in full resolution (which may be higher) for instance for doing Image Processing algorithms on it and in my case upload it to a remote server.

This PR add 2 fields in the response object : 
```typescript
 //only for iOS
 originalImageName?: string;
 originalImagePath?: string;
 ```

### upload large image 
Capture d’écran 2021-01-26 à 14 04 57

### get the original asset (for uploading it to a remote server)
Capture d’écran 2021-01-26 à 14 14 09

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -98,10 +98,17 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
 
 - (void)onImageObtained:(NSDictionary<NSString *,id> *)info {
 
+
+    NSURL *imagePath = [info objectForKey:UIImagePickerControllerReferenceURL];
+    NSString *originalImagePath = imagePath.absoluteString;
+    PHFetchResult *result = [PHAsset fetchAssetsWithALAssetURLs:@[originalImagePath] options:nil];
+    NSString *originalFileName = [[result firstObject] filename];
+
     NSURL *imageURL = [ImagePickerManager getNSURLFromInfo:info];
     self.response = [[NSMutableDictionary alloc] init];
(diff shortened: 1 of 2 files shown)


Johan-dutoit:
Thanks for the PR, looking good.

Couple of things to note
- Readme needs to be updated to include the new values
- Can this be ported to Android (and does it need to be)?

ravirajn22:
Thanks for the effort in making this PR. But I feel the solution you posted has breaking changes, uses deprecated api's and no android support

What do you mean by

> On iOS photos in the picker are limited to 2048x2048px

Why cant you use the fileURI returned already by this library?

Using `fetchAssetsWithALAssetURLs` API requires Photos permission where as currently our library does not require it if using iOS version 11 and above.

`UIImagePickerControllerReferenceURL` is also deprecated.

If you have any other better solution post it here, I am happy to review it.

GaspardC:
@ravirajn22 
> Why cant you use the fileURI returned already by this library?

The fileURI is linked to a downsampled version of large images [https://developer.apple.com/forums/thread/88147](https://developer.apple.com/forums/thread/88147)

> If you have any other better solution post it here, I am happy to review it

ok understood, it fits my usage so I'll keep it like this for now. thanks for your library anyway 👍 

btw in my researches I saw that the new recommended iOS picker `PHPickerViewController` is returning photos in their original dimensions 👌

ravirajn22:
> The fileURI is linked to a downsampled version of large images https://developer.apple.com/forums/thread/88147

Thanks for this info

Yes u can always fork the library and use it the way u want, as u can understand adding features which are not robust will cause maintenance issues over time.

Issue #1428, opened 2020-09-16
Support for PHPicker in iOS 14


emmaielle:
# Feature request

Is support for PHPicker from iOS 14 expected soon?

LeeJunhui:
same question😂

suzushima-ryousuke:
I would also like to know this

davidpaley:
Me too 😄

ravirajn22:
May I know what is the advantage of PHPicker over existing?

ziiw:
Hi,

To answer your question @ravirajn22 it seems that UIImagePickerController have less flexibility.
For instance in UIImagePickerController you can't manage the export options :
https://developer.apple.com/documentation/uikit/uiimagepickercontroller/2890964-videoexportpreset

For me, your package is not usable because the compression phase (after choosing a video), can't be bypassed.

cgarrovillo:
> May I know what is the advantage of PHPicker over existing?

Apple themselves did a whole presentation on it too: https://developer.apple.com/videos/play/wwdc2020/10652/

It holds a lot of features that shouldve been there from the start. I think this is a much necessary change to move forward

jvence:
I think this is a must-have feature

Johan-dutoit:
PR's welcome

ashwinibm:
I'm trying a "limited permission setting" for photos as mentioned in https://github.com/react-native-image-picker/react-native-image-picker/discussions/1564
I'm using react-native-image-picker v3.3.2 and the gallery shows all photos ignoring the limited permissions.

Would https://github.com/react-native-image-picker/react-native-image-picker/pull/1654 support showing only the specific photos that have permission granted on iOS 14+? 
Or is it that permissions aren't required because it'll be a system prompt and we're not really giving the app access with PHPicker?

PR #1654, opened 2021-03-22
Support for `PHPickerViewController`

JanGorman:
## Motivation

This PR adds support for the iOS 14 `PHPickerViewController` to address Issue #1428 

## Test Plan

Tested using the included `example` app to load images/videos and verify that the output matches that of the original.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -8,7 +8,6 @@
 
 @interface ImagePickerManager ()
 
-@property (nonatomic, strong) UIImagePickerController *picker;
 @property (nonatomic, strong) RCTResponseSenderBlock callback;
 @property (nonatomic, retain) NSDictionary *options;
 
@@ -49,23 +48,23 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
     }
     
     self.options = options;
-    self.picker = [[UIImagePickerController alloc] init];
+    UIImagePickerController *picker = [[UIImagePickerController alloc] init];
     
-    [ImagePickerUtils setupPickerFromOptions:self.picker options:self.options target:target];
-    self.picker.delegate = self;
+    [ImagePickerUtils setupPickerFromOptions:picker options:self.options target:target];
+    picker.delegate = self;
     [self checkPermission:^(BOOL granted) {
         if (!granted) {
             self.callback(@[@{@"errorCode": errPermission}]);
             return;
         }
-        [self showPickerViewController];
+        [self showPickerViewController:picker];
     }];
 }
 
-- (void) showPickerViewController {
+- (void) showPickerViewController:(UIViewController *)picker {
     dispatch_async(dispatch_get_main_queue(), ^{
         UIViewController *root = RCTPresentedViewController();
-        [root presentViewController:self.picker animated:YES completion:nil];
+        [root presentViewController:picker animated:YES completion:nil];
     });
 }
 
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -54,25 +54,28 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
     self.options = options;
 
     if (@available(iOS 14, *)) {
-        PHPickerConfiguration *configuration = [ImagePickerUtils makeConfigurationFromOptions:options];
-        PHPickerViewController *picker = [[PHPickerViewController alloc] initWithConfiguration:configuration];
-        picker.delegate = self;
+        if (target == library) {
+            PHPickerConfiguration *configuration = [ImagePickerUtils makeConfigurationFromOptions:options];
+            PHPickerViewController *picker = [[PHPickerViewController alloc] initWithConfiguration:configuration];
+            picker.delegate = self;
 
-        [self showPickerViewController:picker];
-    }
-    else {
-        UIImagePickerController *picker = [[UIImagePickerController alloc] init];
-        [ImagePickerUtils setupPickerFromOptions:picker options:self.options target:target];
-        picker.delegate = self;
-
-        [self checkPermission:^(BOOL granted) {
-            if (!granted) {
-                self.callback(@[@{@"errorCode": errPermission}]);
-                return;
-            }
             [self showPickerViewController:picker];
-        }];
+
+            return;
+        }
     }
+
+    UIImagePickerController *picker = [[UIImagePickerController alloc] init];
+    [ImagePickerUtils setupPickerFromOptions:picker options:self.options target:target];
+    picker.delegate = self;
+
+    [self checkPermission:^(BOOL granted) {
+        if (!granted) {
+            self.callback(@[@{@"errorCode": errPermission}]);
+            return;
+        }
+        [self showPickerViewController:picker];
+    }];
 }
 
 - (void) showPickerViewController:(UIViewController *)picker
@@ -86,10 +89,10 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
 - (void)imagePickerController:(UIImagePickerController *)picker didFinishPickingMediaWithInfo:(NSDictionary<NSString *,id> *)info
 {
     dispatch_block_t dismissCompletionBlock = ^{
-        
         if ([info[UIImagePickerControllerMediaType] isEqualToString:(NSString *) kUTTypeImage]) {
             [self onImageObtained:info];
-        } else {
+        }
+        else {
             [self onVideoObtained:info];
         }
     };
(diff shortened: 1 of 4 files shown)


ios/ImagePickerManager.m
```diff
@@ -3,19 +3,20 @@
...
 #import <AVFoundation/AVFoundation.h>
 #import <Photos/Photos.h>
-#import <React/RCTUtils.h>
+#import <PhotosUI/PhotosUI.h>
 
 @import MobileCoreServices;
 
 @interface ImagePickerManager ()
 
-@property (nonatomic, strong) UIImagePickerController *picker;
```
JanGorman, line 12: The presenting view controller retains the picker so there's no requirement to hold on to it

ios/ImagePickerManager.m
```diff
@@ -198,12 +215,14 @@ - (void)checkCameraPermissions:(void(^)(BOOL granted))callback
     if (status == AVAuthorizationStatusAuthorized) {
         callback(YES);
         return;
-    } else if (status == AVAuthorizationStatusNotDetermined){
+    }
+    else if (status == AVAuthorizationStatusNotDetermined){
```
JanGorman, line 219: There was a mix of formatting

Johan-dutoit:
Thanks for the PR, I'll see if I can squeeze in some testing soon and revert back.

ios/ImagePickerManager.m
```diff
@@ -292,10 +314,114 @@ + (UIImage*)getUIImageFromInfo:(NSDictionary*)info {
...
+                    [picker dismissViewControllerAnimated:YES completion:^{
+                        self.callback(@[response]);
+                    }];
+                });
+            }];
+        }
+    }
+}
+
+- (NSDictionary *)makeResponseFromImage:(UIImage *)image {
```
ravirajn22, line 369: makeResponseFromImage and onImageObtained should be merged into a common function, both perform the same thing but only with different inputs.

ios/ImagePickerManager.m
```diff
@@ -292,10 +314,114 @@ + (UIImage*)getUIImageFromInfo:(NSDictionary*)info {
...
+    }
+
+    response[@"fileName"] = fileName;
+    response[@"width"] = @(image.size.width);
+    response[@"height"] = @(image.size.height);
+
+    return [response copy];
+}
+
+- (NSDictionary *)makeResponseFromURL:(NSURL *)url {
```
ravirajn22, line 407: makeResponseFromURL and onVideoObtained should be merged into a common function.

ravirajn22:
@JanGorman I have added few comments, can you check it and revert with changes or your POV

JanGorman:
@ravirajn22 Thanks for the comments. Agreed! I factored out the common pieces

ios/ImagePickerManager.m
```diff
@@ -269,33 +299,127 @@ - (void)checkPermission:(void(^)(BOOL granted)) callback {
...
+    if (![fileType isEqualToString:@"gif"]) {
+        image = [ImagePickerUtils resizeImage:image
+                                     maxWidth:[self.options[@"maxWidth"] floatValue]
+                                    maxHeight:[self.options[@"maxHeight"] floatValue]];
+    }
+
+    return [self makeResponseFromImage:image fileType:fileType data:data];
+}
+
+- (NSString *)fileTypeFromTypeIdentifier:(NSString *)typeIdentifier
```
JanGorman, line 401: The `typeIdentifier` returns values like `com.compuserve.gif` or `public.png` so doing a bit of fuzzy matching here. There is [UTType](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype) but that's for Swift only

ravirajn22:
Need to check the PR again, will update and push before this weekend.

Johan-dutoit:
I did some testing and mostly works okay.

I get the following error when quickly tapping an image twice

image

JanGorman:
@Johan-dutoit Thanks for the feedback. I'll see how to work around that

ios/ImagePickerManager.m
```diff
@@ -269,33 +299,126 @@ - (void)checkPermission:(void(^)(BOOL granted)) callback {
...
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [picker dismissViewControllerAnimated:YES completion:^{
+                self.callback(@[@{@"didCancel": @YES}]);
+            }];
+        });
+        return;
+    }
+
+    // Immediately dismiss the controller to prevent calling the completion handler twice
+    [picker dismissViewControllerAnimated:YES completion:nil];
```
JanGorman, line 350: @Johan-dutoit This change here fixes the double tap issue in my tests

Johan-dutoit: It does a better job, but still fails if you click really fast.

JanGorman: You really want to try to break it 😅 I would suggest you make a call on how realistic this is (double tap speed vs. mouse clicks…). I don't see what other way there is to prevent this.

What's interesting is that it's reproducible (at least for me) only on the first double click. Subsequent attempts fail until you relaunch it. I'm able to reproduce this in a bare bones example and will report it as a bug to Apple.

```swift
class ViewController: UIViewController {
    @IBAction func pick(_ sender: Any) {
        let picker = PHPickerViewController(configuration: .init())
        picker.delegate = self
        present(picker, animated: true)
    }
}

extension ViewController: PHPickerViewControllerDelegate {
    func picker(_ picker: PHPickerViewController, didFinishPicking results: [PHPickerResult]) {
        picker.dismiss(animated: true)
        // Prints twice the first time you double click really quickly
        print("Called")
    }
}
```

JanGorman: Reported to Apple as `FB9059292`

Johan-dutoit: I totally agree with your sentiments above.  At first I didn't even want to mention it, but decided that it's good to have a trail.

Johan-dutoit:
if @ravirajn22 is happy with the code, I'm happy to give the thumbs up.

Sven65:
Does this PR allow users of the library to set the [selection limit](https://developer.apple.com/documentation/photokit/phpickerconfiguration/3606592-selectionlimit), in order to select multiple images at once?

JanGorman:
@Sven65 It does not. The library currently doesn't offer this option either. It's straightforward to add that option but IMO better done in a separate PR

JanGorman:
@ravirajn22 Have you had  chance to review? If there's anything I can do to help unblock the PR please let me know 🙂

ios/ImagePickerManager.m
flochtililoch: Shouldn't the call to `dismissViewControllerAnimated` be made before we check for results? That's what's suggested [in the WWDC video about PHPicker](https://developer.apple.com/videos/play/wwdc2020-10652/?time=392), I'm wondering if that has incidence in the bug you're facing with double/triple taps.

JanGorman: @flochtililoch Thanks for the comment. I made this change specifically to attempt to work around the mentioned bug. In my tests it didn't matter whether you call the completion block in the dismiss completion block or outside. It was always reproducible the _first_ time that you present the picker but not on the second call which to me suggest that there is an issue.

flochtililoch: Oh, what I meant was:

```objc
    // First, dismiss picker
    [picker dismissViewControllerAnimated:YES completion:nil];

    // Then, check results
    if (results.count == 0) {
        dispatch_async(dispatch_get_main_queue(), ^{
            [picker dismissViewControllerAnimated:YES completion:^{
                self.callback(@[@{@"didCancel": @YES}]);
            }];
        });
        return;
    }
    ...
```

Another thing worthwhile to explore, is around setting the flag `preferredAssetRepresentationMode` on the configuration object to `PHPickerConfigurationAssetRepresentationModeCurrent`. This [blog article](https://www.appcoda.com/phpicker/#crayon-60825b230a088198413804) suggests that not setting it can have a performance impact on the import process.

Johan-dutoit: Let's give this a go, would love to now face this :D

JanGorman: I moved it to the first line of the delegate callback. The double tap problem persists but in terms of consistency I guess this is better (only one dismiss instead of separate handling for the "cancel" case).
I also applied the suggestion to set the `preferredAssetRepresentationMode`

Johan-dutoit:
@ravirajn22  last chances for a review, going to merge this today, unless there are any objecetions.

ravirajn22:
@Johan-dutoit I will review and tell before tomorrow. This change affects many lines including styling and refactor of unrelated lines. I don't know if its the right way to add feature + refactor together. Just give some time.

ravirajn22:
> I did some testing and mostly works okay.
> 
> I get the following error when quickly tapping an image twice
> 
> image

Regarding this error, we can handle this in another PR. Just setting the callback to nil after sending response and checking if callback != nil before sending the response should fix this issue. This issue exist even in current production build I think.

ios/ImagePickerManager.h
```diff
@@ -1,11 +1,12 @@
...
 #import <UIKit/UIKit.h>
+#import <PhotosUI/PhotosUI.h>
 
 typedef NS_ENUM(NSInteger, RNImagePickerTarget) {
   camera = 1,
   library
 };
 
-@interface ImagePickerManager : NSObject <RCTBridgeModule, UINavigationControllerDelegate, UIActionSheetDelegate, UIImagePickerControllerDelegate>
+@interface ImagePickerManager : NSObject <RCTBridgeModule, UINavigationControllerDelegate, UIActionSheetDelegate, UIImagePickerControllerDelegate, PHPickerViewControllerDelegate>
```
JanGorman, line 10: @ravirajn22 Out of curiosity, why do you prefer doing this on the public interface? For one, the outside world doesn't need to know that it conforms to the interface (personally I'd move `UIActionSheetDelegate`, `UIImagePickerControllerDelegate`, `UINavigationControllerDelegate` into the `.m` file too). This was one of the benefits that Obj-C has over Swift, that you can move the `#import` into the `.m` file making the link phase faster

ios/ImagePickerManager.m
```diff
@@ -111,40 +108,58 @@ - (void)imagePickerControllerDidCancel:(UIImagePickerController *)picker
...
+        else if ([provider hasItemConformingToTypeIdentifier:(NSString *)kUTTypeMovie]) {
+            [provider loadFileRepresentationForTypeIdentifier:(NSString *)kUTTypeMovie
+                                            completionHandler:^(NSURL * _Nullable url, NSError * _Nullable error) {
+                [self onVideoObtained:url];
+            }];
+        }
+    }
+}
 
+- (void)onImageObtained:(UIImage*)image data:(NSData*)data
```
JanGorman, line 139: Come to think of it, we don't necessarily need to pass the `UIImage` here. We could
```objective-c
UIImage *image = [[UIImage alloc] initWithData:data];
```

ios/ImagePickerManager.h
ravirajn22: I am no expert in objective c, so l like consistency over others. Since all previous protocols have been done on the public interface as you said, I thought its better if we can declare PHPickerViewControllerDelegate the same way.

Refactors can go in another PR or with smaller PR, not with big PR like this.

JanGorman: Understood! Then I'll gladly change this in a follow-up PR 👍

ios/ImagePickerManager.m
JanGorman: Fine to put this into a follow-up PR as well imo

ravirajn22: Yes ideally it should be done as you said. But in UIImagePickerController there is something called edited UIIImage and original UIImage. But the data we obtain is only from original image, you cannot get edited image data. So if the user edited the image then we need to process further using the edited UIImage and use the data from original image to just obtain image type and save it directly for GIF type. I am not sure if I am explaining properly.

This is how previous code has been working so I just used it.

JanGorman: I'll look into it. Should be possible to figure out the type and do the editing as you said

JanGorman:
LGTM

Johan-dutoit:
Thanks @JanGorman, I'll do some basic testing locally and get this merged.

ios/ImagePickerManager.m
```diff
@@ -224,6 +224,9 @@ - (void)onVideoObtained:(NSURL *)url
     }
     NSDictionary *response = @{@"uri": videoDestinationURL.absoluteString};
 
+    AVAsset *asset = [AVAsset assetWithURL:videoDestinationURL];
+    self.response[@"duration"] = @(roundf(CMTimeGetSeconds(asset.duration)));
+    self.response[@"uri"] = videoDestinationURL.absoluteString;
```
Johan-dutoit:
```
ios/ImagePickerManager.m:228:10: Property 'response' not found on object of type 'ImagePickerManager *'
ios/ImagePickerManager.m:229:10: Property 'response' not found on object of type 'ImagePickerManager *'
```

JanGorman: right, need to update those after the merge. I'll update it sometime later this afternoon and ping you then. Sorry for the hassle @Johan-dutoit

Johan-dutoit: I made the change, just check if you're happy.

JanGorman: amazing, thanks

ios/ImagePickerManager.m
```diff
@@ -222,11 +222,14 @@ - (void)onVideoObtained:(NSURL *)url
           }
         }
     }
-    NSDictionary *response = @{@"uri": videoDestinationURL.absoluteString};
+    
+    NSMutableDictionary *response = [[NSMutableDictionary alloc] init];
+    response[@"uri"] = videoDestinationURL.absoluteString;
```
ravirajn22, line 227: This line is duplicate of line 232.

Johan-dutoit: Good catch, bad copy paste!

Johan-dutoit: Sorted

JanGorman:
@Johan-dutoit Is there anything left that I can help with to merge? :)

Johan-dutoit (approved):

gabrieldonadel:
Does this support the `durationLimit` param?

PR #1669, opened 2021-04-19
docs(readme): updated iOS Response object fileName property

yerevin:
## Motivation

I ran into issue with fileName being undefined on iOS on videos when tried to parse it
fileName property is available only for Android photos/videos and photos in iOS, on iOS video fileName is undefined.

## Test Plan

This is only README update so I don't think any test plan is needed


Johan-dutoit:
Is there not a code change needed instead, to support it for iOS videos?

yerevin:
Code change can be better solution for this, but I don't know Objective-C, so decided to at least "help" in this way :)

Johan-dutoit:
Fair enough! If you can raise an issue as well please, then ideally someone that is familiar with Objective-C can take a look.
This can then get merged in until a fix is applied.

yerevin:
#1670

ravirajn22:
Thanks for putting effort to open a PR. This will be fixed in library.

PR #1671, opened 2021-04-22
fix(android): read minSdkVersion from root project

keremoge:


ravirajn22:
Thanks for putting effort to open a PR. 
But this library requires min 21 to work properly so there is no use in bringing this value from root project. But you can in ur main project have any version above this version.

PR #1693, opened 2021-05-06
feat: add prop "copyTo" to have a copy of the file in cache folder or document folder

gcoro:
## Motivation (required)

Hi, I added the prop 'copyTo', similar as how it's done in [this library](https://github.com/rnmods/react-native-document-picker), to allow the user to store the photo/video in the cache directory or document directory. 
This because, using the tmp directory in iOS, I had a problem where the files got deleted from the directory before I could actually upload them to the server. 
If copyTo is specified, the copied file uri is returned in the response as ‘fileCopyUri’, otherwise this field just contains the same original uri.

## Test Plan (required)

I tested with the picker on ios and android.


ravirajn22:
Thanks for your effort, but use any other library to move files to private storage. We wont support this for now.

PR #1711, opened 2021-05-18
feat(android): change launch image library intent type to ACTION_GET_CONTENT

norbertsongin:
https://developer.android.com/guide/topics/providers/document-provider#client


norbertsongin:
Created by mistake, can be removed

PR #1713, opened 2021-05-19
Make REQUEST_LAUNCH codes public

kas:

## Motivation (required)

Thanks in advance for taking a look. The product I work on uses react-native-image-picker. We would like to hook directly into these REQUEST_LAUNCH codes, but newer versions of this package don't expose them publicly anymore. I looked at the change that took them from public to private, and there didn't seem to be an explicit reason for the change. This change just makes them public again.

## Test Plan (required)

Just a minor change to make it easier for consuming apps to use these codes. Just making the 4 codes public.


ravirajn22:
Tell us how this hook works, why do you require request codes? The best thing would be just fork this library, make changes and use it. Since this change is simple we might consider it, but I don't think its right.

Johan-dutoit (approved):

PR #1706, opened 2021-05-17
feature: add ability to pick multiple images/ videos

rpander93:
## Motivation

This PR adds the ability to select multiple images/ videos. This feature has been requested in the past but was rejected due to no native iOS support. Now that the library uses PhotoKit on iOS, the native support is there.

## Test Plan

Tested this functionality on iOS simulator, iPhone 11 & Nexus 6P running Android 8.

_Note: made some changes to the example app but couldnt get it running otherwise._

PR #1726, opened 2021-05-27
Fix issue in multiple file selection

ravirajn22:
- Fixed the resize issue in android
- Fixed gif image issue in iOS
- Also rename `allowMultiple` to `selectionLimit` since iOS supports providing an integer value
- Updated Readme

@Johan-dutoit @rpander93 Please review and test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ A React Native module that allows you to select a photo/video from the device li
   <img src="https://img.shields.io/npm/v/react-native-image-picker" />
 </p>
 
-## Make sure you'r reading the doc applicable to your version, for example if your using version 3.8.0 go to tag 3.8.0 and read those docs. This doc is always that of master.
-## Also read version release notes for any breaking changes especially if you'r updating the major version.
+### Make sure you'r reading the doc applicable to your version, for example if your using version 3.8.0 go to tag 3.8.0 and read those docs. This doc is always that of master.
+### Also read version release notes for any breaking changes especially if you'r updating the major version.
 
 
 # Install
@@ -86,7 +86,7 @@ The `callback` will be called with a response object, refer to [The Response Obj
 | cameraType    | OK  | OK      | 'back' or 'front'. May not be supported in few android devices                                        |
 | includeBase64 | OK  | OK      | If true, creates base64 string of the image (Avoid using on large image files due to performance)     |
 | saveToPhotos  | OK  | OK      | (Boolean) Only for launchCamera, saves the image/video file captured to public photo                  |
-| selectionLimit| OK  | OK      | Default is 1, 0 to allow any number of files. Only iOS version >= 14 support `0`|
+| selectionLimit| OK  | OK      | Default is `1`, use `0` to allow any number of files. Only iOS version >= 14 support `0` and also it supports providing any integer value|
 
 ## The Response Object
 
--- a/README.md
+++ b/README.md
@@ -7,13 +7,9 @@ A React Native module that allows you to select a photo/video from the device li
   <img src="https://img.shields.io/npm/v/react-native-image-picker" />
 </p>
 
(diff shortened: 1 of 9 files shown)


ios/ImagePickerManager.m
```diff
@@ -89,17 +89,14 @@ - (void)imagePickerController:(UIImagePickerController *)picker didFinishPicking
 
         if ([info[UIImagePickerControllerMediaType] isEqualToString:(NSString *) kUTTypeImage]) {
             UIImage *image = [ImagePickerManager getUIImageFromInfo:info];
-            [assets addObject:[self mapImageToAsset:image data:nil]];
+            [assets addObject:[self mapImageToAsset:image data:[NSData dataWithContentsOfURL:[ImagePickerManager getNSURLFromInfo:info]]]];
```
rpander93, line 92: `nil` is passed in intentionally. the key getNSURLFromInfo tries to get doesnt exist. see my comment (and debugger output): https://github.com/react-native-image-picker/react-native-image-picker/pull/1706/files#r633383052

ios/ImagePickerManager.m
```diff
@@ -89,17 +89,14 @@ - (void)imagePickerController:(UIImagePickerController *)picker didFinishPicking
...
         } else {
             NSDictionary *asset = [self mapVideoToAsset:info[UIImagePickerControllerMediaURL]];
-            if (asset == nil) return;
```
rpander93, line 95: mapVideoToAsset can fail. in that case it will have called the callback already with the error.

now, it will add a nil video to the assets array and later call the callback again (which will throw because it can only be called once)

rpander93 (commented):
looking quite good. i intentionally named it `allowMultiple` bc i took Android as the lowest common denominator. but this is also fine to me.

ios/ImagePickerManager.m
ravirajn22: Did u check in ios13, because this value cannot be nil according to ios docs

ios/ImagePickerManager.m
ravirajn22: Ya, just now checked then we must refactor that function to throw error instead of calling callback and catch it where we call.

ios/ImagePickerManager.m
rpander93: the screenshot is from a ios 14.5 sim, although i cannot reproduce it on current branch. so seems all good to me. (it will work regardless)

ios/ImagePickerManager.m
rpander93: indeed, seems good. didnt want to include that many changes in my original PR.

rpander93 (dismissed):

README.md
```diff
@@ -7,13 +7,9 @@ A React Native module that allows you to select a photo/video from the device li
   <img src="https://img.shields.io/npm/v/react-native-image-picker" />
 </p>
 
-**Note:** If you are still using deprecated version 2.x.x [check](https://github.com/react-native-image-picker/react-native-image-picker/tree/2.3.4) this for documentation.
+### Make sure you'r reading the doc applicable to your version, for example if your using version 3.8.0 go to tag 3.8.0 and read those docs. This doc is always that of master.
```
Johan-dutoit, line 10: `you're` and `master` to be renamed to `main`

README.md
```diff
@@ -7,13 +7,9 @@ A React Native module that allows you to select a photo/video from the device li
...
+### Also read version release notes for any breaking changes especially if you'r updating the major version.
```
Johan-dutoit, line 11: `you're`

Johan-dutoit:
FWIW I marked 4.0.0 as a beta on npm... latest is still 3.8.1

ios/ImagePickerManager.m
ravirajn22: refactored

ravirajn22:
@Johan-dutoit done.

Issue #1742, opened 2021-06-08
[🐛] App crashed after select Video from Library in iOS 13


vivianlys90:
### Description

Getting this error after select Video from Library in iOS 13. App crashed at function didFinishPickingMediaWithInfo.

"Error returned from daemon: Error Domain=com.apple.accounts Code=7 "(null)""

<img width="1127" alt="Screenshot 2021-06-08 at 10 29 11 AM" src="https://user-images.githubusercontent.com/22317616/121113476-aebec000-c844-11eb-8c54-00affdf0cbfd.png">

### How to repeat issue and example

- Use this options for Video selection:
{
    "durationLimit": 600,
    "maxHeight": 1000,
    "maxWidth": 1000,
    "mediaType": "video",
    "quality": 1,
    "videoQuality": "low"
}
- Choose Video from Library
- Showing Video viewer screen
- Tap on 'Choose' and compress video
- App crashed

### Additional Information

- Image Picker version: 4.0.3
- React Native version: 0.63
- Platform: iOS
- Development Operating System: MacOS
- Dev tools: Xcode 12.4

m3rlin94:
Happening this to me as well.

m3rlin94:
Can confirm this is happening on v4.0.4

vivianlys90:
Hi Naveen,
I have submitted my fixes and now still waiting for code review.You can check this pull request #1743
Have a good day.
Best Regards,Vivian Lee


    On Tuesday, 29 June 2021, 09:42:29 am MYT, Naveen Cooray ***@***.***> wrote:  
 
 


Happening this to me as well.

m3rlin94:
> Hi Naveen, I have submitted my fixes and now still waiting for code review.You can check this pull request #1743 Have a good day. Best Regards,Vivian Lee On Tuesday, 29 June 2021, 09:42:29 am MYT, Naveen Cooray ***@***.***> wrote: Happening this to me as well. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Thanks @vivianlys90! Will check that out

Muffles:
Same here, can confirm this is happening on 4.0.4

klakshman8:
I am still having issue any solution

vivianlys90:
I have submitted PR for this issue, but still waiting for code merge. #1743

musesfish:
when I picked a video in Android 
APP crashed in recently
react–native 0.59.1
[...]

Issue #1687, opened 2021-05-02
[🐛] Getting Issue in image selection on Macbook Pro M1


thejogsa:
### Description

```
ImagePicker.showImagePicker(options, (response) => {

            if (response.didCancel) {
                console.log('User cancelled image picker');
            } else if (response.error) {
                console.log('ImagePicker Error: ', response.error);
            } else if (response.customButton) {
                console.log('User tapped custom button: ', response.customButton);
            } else {
                const source = response;
                setImage(source)
            }
        });
```

<img width="1105" alt="Screenshot 2021-05-02 at 7 38 24 PM" src="https://user-images.githubusercontent.com/8350797/116816034-f97e5580-ab7d-11eb-9a36-f620efd33af8.png">

```
```

### Error Getting

```
2021-05-02 19:38:05.445599+0530 PlayTubeVideo[5424:116727] [xpc.exceptions] <NSXPCConnection: 0x600001c146e0> connection on anonymousListener or serviceListener from pid 5954: Warning: Exception caught during invocation of selector didSelectMediaWithInfoDictionary:, dropping incoming message and invalidating the connection.

Exception: *** -[NSURL URLByAppendingPathExtension:]: component, components, or pathExtension cannot be nil.
*** -[NSURL URLByAppendingPathExtension:]: component, components, or pathExtension cannot be nil.
```

```
```

### How to repeat issue and example

- When selecting image on  Macbook Pro M1 in iOS Simulator. Getting this error.
- ..

### Solution

What needs to be done to address this issue?

### Additional Information

- Platform: [FILL THIS OUT: iOS Simulator ]
- Development Operating System: [FILL THIS OUT: Macbook Pro M1]
- Dev tools: [FILL THIS OUT: Xcode 12.4 and Mac OS Big Sur ]

Johan-dutoit:
Can you assist in debugging and fixing this?  I don't believe any of the current maintainers have an M1 mac to fix.

phillip9077:
I have an M1 mac. I'm facing a similar issue of not being able to select an image, though my issue stems from the iOS simulator not being able to detect image selection at all. I can get to the point as you see in the screenshot, but when I press on any of the pictures nothing happens. The cancel, photos, and albums button responds fine though. 

[...]

Issue #1647, opened 2021-03-18
Callback doesn't return anything [🐛] 


JosephMulder:
### Description
 I'll have launchImageLibrary and launchCamera running and they open the photo picker or the camera, but once I select the photo and return to the application the callback doesn't run. 

### Solution

What needs to be done to address this issue?

### Additional Information

Additional Information
Image Picker version: ["react-native-image-picker": "^3.3.2"]
React Native version: ["react-native": "0.63.4",]
Platform: [Android]

ravirajn22:
Show your code, how do u use. Show javascript and android logs.

JosephMulder:
> Show your code, how do u use. Show javascript and android logs. Just think, how do I diagnose with the data you provided?

Sorry, updated

ravirajn22:
Strange there is no issues in the latest version I check. If possible can you check with the same code in iOS and see if callback is called? BTW there is not customButton option in latest version, check the docs carefully.

habeebtheprogrammer:
i'm having same issue.

rdjuric:
Same here. Started happening after I updated from 2.xx and removed the permissions from the manifest.
I've only been able to reproduce it on Android 28 (the callback is correctly called on Android >= 29 & iOS).

I'm at
"react-native": "0.61.5",
"react-native-image-picker": "^3.3.2"

ViktorVojtek:
Same here. iOs 14.5.

React native 0.63.4

Callback with response is not triggering from version ˆ3.3.2
Version 3.3.1 works fine on real device, but can't pick image from library on simulator.

HarounAns:
Im having the same issue. @ViktorVojtek do you have an M1. Mine is an M1, not sure if that could potentially be the issue:

```
import React, { Component } from 'react';
import { Text, SafeAreaView, TouchableOpacity } from 'react-native';
import { launchImageLibrary } from 'react-native-image-picker';


export default class App extends Component {
    handleChoosePhoto = () => {
        console.log('handleChoosePhoto');

        const options = {};
        launchImageLibrary(options, response => {
            console.log('response');
            console.log(response);
        })
    }

    render() {
        return (
            <SafeAreaView>
                <TouchableOpacity onPress={this.handleChoosePhoto}>
                    <Text>Choose Photo</Text>
                </TouchableOpacity>
            </SafeAreaView>
        )
    }
```

[...]

PR #1743, opened 2021-06-08
fix: App crashed after select Video from Library in iOS 13 (#1742)

vivianlys90:
## Motivation

Fix issue for App crashed after select Video from Library in iOS 13. #1742 

I made the code changes by referring to react-native-image-picker v2.3.4
- Create a new object `NSError * err = nil;` to replace error checking in `mapVideoToAsset`
- The new `err` will use to detect error returned from `([fileManager isWritableFileAtPath:[url path]])`

## Test Plan
1. Use iOS 13 device to open react-native-image-picker.
2. Select Video from image picker.
3. Should able to get video without crashing.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -248,7 +248,7 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
             [fileManager copyItemAtURL:url toURL:videoDestinationURL error:error];
           }
 
-          if (*error) {
+          if (error && *error) {
               return nil;
           }
         }
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -93,7 +93,7 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
             [assets addObject:[self mapImageToAsset:image data:[NSData dataWithContentsOfURL:[ImagePickerManager getNSURLFromInfo:info]]]];
         } else {
             NSError *error;
-            NSDictionary *asset = [self mapVideoToAsset:info[UIImagePickerControllerMediaURL] error:nil];
+            NSDictionary *asset = [self mapVideoToAsset:info[UIImagePickerControllerMediaURL] error:&error];
             if (asset == nil) {
                 self.callback(@[@{@"errorCode": errOthers, @"errorMessage":  error.localizedFailureReason}]);
                 return;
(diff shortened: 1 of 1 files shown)


ios/ImagePickerManager.m
```diff
@@ -241,15 +241,15 @@ -(NSMutableDictionary *)mapVideoToAsset:(NSURL *)url error:(NSError **)error {
         }
 
         if (url) { // Protect against reported crash
-
+            NSError * err = nil;
```
ravirajn22, line 244: `mapVideoToAsset:error` function already takes an NSError(error) as a param. We use this `error` in [FileManager...] call. So why do we require this change.

ravirajn22: The place we call `mapVideoToAsset:error` function expects the error object to be set if there is any error, here ur creating local variable which will not be reflected.

vivianlys90: Hi @ravirajn22 ,

Thanks for the prompt review. After debug again and I found out my fixes is not efficient. 

Actually I'm trying to bypass the error pointer passed from 
`- (void)imagePickerController:(UIImagePickerController *)picker didFinishPickingMediaWithInfo:(NSDictionary<NSString *,id> *)info` to `mapVideoToAsset`, and only do checking for the FileManager error. 

The reason I do that is because I found out after done picking Video file from image picker, the error is nil. But after passed to 
`mapVideoToAsset` the error pointer is holding something. And end up returned nil error value for callback.

You can refer to my video for more details.

https://user-images.githubusercontent.com/22317616/121286443-6f5ea500-c912-11eb-8fd9-9beb6cc87fda.mov

vivianlys90:
Hi @ravirajn22 

I have commit a new changes. Now will check (*error) in mapVideoToAsset function.

The logic is like this:
NSError **error is a pointer, so checking on 
`if (error) {`
not make sense because it is refer to the instance and it will always have value.

So we need to change to (*error) to check the pointer object.

vivianlys90:
Hi any update?

vivianlys90 (commented):
Change to `if (*error)` to check pointer object value.

ravirajn22:
@vivianlys90 give some time, need to test manually.

klakshman8:
after selecting video nothing is returning

vivianlys90:
> after selecting video nothing is returning

Hi, your app crashed after select video or no callback returned?

klakshman8:
no call back

klakshman8:
I tried this solution but did not work crashing on android

klakshman8:
any solution please have been fighting with it from last few days

vivianlys90:
> I tried this solution but did not work crashing on android

This solution is for iOS, should not impact android. And the no callback issue suspect is due to macbook M1 chip. #1687, #1647

klakshman8:
yep
my mac is macbook pro m1 chip
but previous version working fine

batuyildiz-bv:
Can confirm that this PR solves the issue for us too.

lukebrandonfarrell:
Closing this as we have a PR on the way which will fix it.

tdbrian:
See PR https://github.com/react-native-image-picker/react-native-image-picker/pull/1904

PR #1770, opened 2021-07-08
[Android] Add Safe SDK Version control

eomttt:

## Motivation (required)

When i install image-picker and install android error like this
스크린샷 2021-07-08 오후 11 56 06

so i tried find another projects, almost they have get safeExtGet function.
so i tried this module in locally, then it works very well!!!

I don't know exactly fallback version. but this safe function solved my problem.
please write exactly fallback version

## Test Plan (required)

A good test plan has the exact commands you ran and their output, provides screenshots or videos if the pull request changes UI or updates the website. See [What is a Test Plan?][1] to learn more.  

If you have added code that should be tested, add tests.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,11 +1,15 @@
 apply plugin: 'com.android.library'
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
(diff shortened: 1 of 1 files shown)


lukebrandonfarrell:
Not sure if this actually solves a problem, I think this is most likely caused by your own project configuration.

PR #1762, opened 2021-06-23
fix: HECI image issue for ios

nomi9995:
Thanks for submitting a PR! Please read these instructions carefully:

- [x] Explain the **motivation** for making this change.
- [x] Provide a **test plan** demonstrating that the code is solid.
- [x] Match the **code formatting** of the rest of the codebase.
- [x] Target the `master` branch, NOT a "stable" branch.

## Motivation (required)

when we choose the `HEIC` image from the library then this library makes the image in tmp directory but in `HEIC` it is not making an image in tmp directory. I added support for `HEIC` image

fixed #872 & #890

## Test Plan (required)
I have tested this on my phone. There are no automated tests in the repo.
you can choose HEIC image from the library then you will see the `HEIC` image is making in tmp directory

Issue #890, opened 2018-07-13
ios Heic  uploading error


252405649:
### Description

type heic Picture uploading

error:
nested exception is java.lang.RuntimeException: net.coobird.thumbnailator.tasks.UnsupportedFormatException: No suitable ImageWriter found for HEIC.] with root cause

### How to repeat issue and example

```
ImagePicker.launchImageLibrary(options, (response) => {
      if (response.didCancel) {
        console.log('User cancelled image picker');
      }
      else if (response.error) {
        console.log('ImagePicker Error: ', response.error);
      }
      else {
        let source = {}
          source = {uri: response.uri, type: response.type, name: response.fileName}
          this.handleChange(source)
          this.setState({
            loading: false
          });
     }
 });  

```
### Solution
react-native-heic-converter

### Additional Information

* React Native version: 0.55.4
* Platform: iOS
* Development Operating System: MacOS
* Dev tools: iOS

252405649:
```
import RNHeicConverter from 'react-native-heic-converter';

ImagePicker.launchImageLibrary(options, (response) => {
     if (response.didCancel) {
        console.log('User cancelled image picker');
      }
      else if (response.error) {
        console.log('ImagePicker Error: ', response.error);
      }
      else {
        let source = {}
        if (Platform.OS === 'ios' && (response.fileName.endsWith('.heic') || response.fileName.endsWith('.HEIC'))) {
          RNHeicConverter.convert({ path: response.origURL }).then((data) => {
            const { success, path, error } = data
            if(!error && success && path) {
              let name = response.fileName
              name = name.replace(".heic", ".jpg")
              name = name.replace(".HEIC", ".jpg")
              source = {uri: path, name}
              this.handleChange(source)
              this.setState({
                loading: false
              });
            } else {
              Toast.show({text: "heic类型转jpg失败!"});
            }
          });
        } else {
          source = {uri: response.uri, type: response.type, name: response.fileName}
          this.handleChange(source)
          this.setState({
            loading: false
          });
        }
      }
    });
```

MrBrownser:
Hi, leaving it here for reference as well. Is not necessary to install another module, just grab always the `uri` from the response and update the filename in case you are having a _heic_ image. Code example:

```
const options = {
    title: 'Select Avatar',
    storageOptions: {
    skipBackup: true,
    path: 'images'
    },
    noData: true,
    quality: 1,
    mediaType: 'photo'
};

[...]

Issue #872, opened 2018-06-19
Heic format


dlochynski:
Is there any way to deal with .heic images selection on the iPhones? I can't currently even pick them.

b-d055:
Hi @dlochynski I got around this problem by noticing that all images have an uri (response.uri) that ends with .jpg. Even those images with a .heic file format and fileName. I simply use this .jpg uri when I want to do something with the image, including uploading to an API endpoint.

dlochynski:
I must've ommited that. It was quite confusing for me that the format was different than the response.uri. Thanks a lot.

AlMel3000:
Hi!
Somewhy iOS converts HEIC to jpg and MOV to mp4
So as I need uncompressed and uncchanged file I did so:

`    _selectImage() {

        const options = {
            noData: true,
            quality: 1,
            mediaType: 'mixed',
            videoQuality: 'high',
            permissionDenied: {
                title: LocalisationHandler.getAppLocale().permission_denied_title,
                text: LocalisationHandler.getAppLocale().permission_denied_message

            }
        };

        ImagePicker.launchImageLibrary(options, (response) => {
            console.warn('ImagePicker RESPONSE = ', response);

            if (response.didCancel) {
                console.warn('ImagePicker User cancelled image picker');
            }
            else if (response.error) {
                console.warn('ImagePicker Error: ', response.error);
            }
            else if (response.customButton) {
                console.warn('ImagePicker User tapped custom button: ', response.customButton);
            }
            else {
                let path;
                let uri = response.uri;
                let origURL;
                let imageName = response.fileName;
                if (Platform.OS === 'ios') {
                    path = null;
                    origURL = response.origURL;
                    isIosImage = true;
                } else {
                    path = response.path;
                }
                this.processFile(path, uri, imageName, origURL)
            }
        });

    }

    processFile(path, uri, fileName, origURL) {
        if (path !== null) {
            RNFS.hash(path, 'sha512')
                .then((result) => {
                    this.doSomeWork(result, fileName);
                });
        } else {
            RNFS.copyFile(uri, `${RNFS.TemporaryDirectoryPath}${fileName}`);
            RNFS.copyFile(origURL, `${RNFS.TemporaryDirectoryPath}${fileName}`);
            path = `${RNFS.TemporaryDirectoryPath}${fileName}`;
            console.warn('path : ', path);

            RNFS.hash(path, 'sha512')
                .then((result) => {
                    this.doSomeWork(result, fileName);
[...]

PR #1788, opened 2021-08-02
docs: update docs for HEIC image for simulator

nomi9995:

## Motivation (required)

when we choose the `HEIC` image from the library then this library makes the image in tmp directory but in `HEIC` it is not making an image in tmp directory but I have learned it does not work only on the simulator

#1762 and #1787 were hacky solutions therefore we should not merge those PR.

#872 & #890

## Test Plan (required)

I have tested this on my phone. There are no automated tests in the repo.
when you will choose HEIC image from simulator then it will not work but if you will choose HEIC image from device then it will work

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,12 @@ Image/video captured via camera will be stored in temporary folder so will be de
 | permission         | Permission not satisfied                          |
 | others             | other errors (check errorMessage for description) |
 
+## Known issues
+
+### iOS
+**HEIC:** Starting with version **0.8.1** the iOS implementation uses PHPicker to pick (multiple) images on iOS 14 or higher.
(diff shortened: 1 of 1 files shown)


afilp:
Is there a problem with this PR? Our app crashes when HEIC is selected and going to be uploaded.

nomi9995:
@afilp are you testing on simulator?

afilp:
@nomi9995 Not in the simulator, in a real device, but now I am not sure if this is related to HEIC. It is for sure that Android users with HEIC upload images that seem to be "jpg" but they are not! They are still HEIC inside, as they do not open in a Windows PC or a website but these "jpg" open in a MAC. Also they have the "HEIC" letters inside their binary data.
Is there a solution for that?

Issue #1749, opened 2021-06-11
[🐛] Limited permission not work for iOS


nicolasdevienne:
Limited permission not work for iOS. When I ask user to choose permission to access Library, he can choose all photos or just 1, 2 or 3 photos for example. When he choose just 1, 2 or 3 photos, just after, launchImageLibrary shows all photos in Library. No filter is apply.

Can you help me ?

jeeyunlee-fe:
same issue here.

hoony11:
same issue here. 

https://github.com/react-native-image-picker/react-native-image-picker/blob/1878acbca5ca40873cdbe047869f2e6337b28b49/ios/ImagePickerManager.m#L283

iOS 14 Limited permission not support

tomgransden:
I get the permission dialog after taking a photo with saveToPhotos: true. Selecting some photos has no effect the next time I use launch the image library.

ndevienne:
react-native-image-crop-picker works perfectly for iOS ! :)

hazyikmis:
-> Has anybody found any workaround for this?

(Sorry but this library is unusable for ios without fixing this issue.)

thibmaek:
Did my best Obj-C effort to integrate limited photo support here: https://github.com/react-native-image-picker/react-native-image-picker/pull/1814

MusabBasheer:
any update

cancerberoSgx:
Any updates on this issue ? The library is awesome but this issue is critical for our users. Thanks

thatgriffith:
I'm also looking for this feature. Any updates? 🙂

gurbela:
any news?

thatgriffith:
We decide to use https://github.com/ivpusic/react-native-image-crop-picker instead. Works like a charm!

FadiAboMsalam:
Any suggestion for this ????

thatgriffith:
> Any suggestion for this ????

We switched to https://github.com/ivpusic/react-native-image-crop-picker and it works fine now.

jeromegrosse:
Any update on that other than changing library? `react-native-image-crop-picker` have issues on its own such as GIF support on iOS.

navaspavil:
Any update on this?

fellipesnoww:
Any update on this?

Johan-dutoit:
Would happily review PRs that solve this.

Johan-dutoit:
[...]

PR #1814, opened 2021-09-07
Support limited photo picking

thibmaek:
## Motivation (required)

This is an aim to solve issue https://github.com/react-native-image-picker/react-native-image-picker/issues/1749

I added a check for `PHAuthorizationStatusLimited` and presented a limited picker view controller. My Obj-C is pretty rusty/non-existing so this is the best effort I could do. If it requires changes please let me know so I can fix them.

## Test Plan (required)

On an iOS 14+ device, open the image library and select limited photo permissions.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -63,6 +63,12 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
 
 #if __has_include(<PhotosUI/PHPicker.h>)
     if (@available(iOS 14, *)) {
+        PHAuthorizationStatus status = [PHPhotoLibrary authorizationStatus];
+        if (status == PHAuthorizationStatusLimited) {
+            [self showLimitedPickerViewController];
+            return;
+        }
+    
         if (target == library) {
             PHPickerConfiguration *configuration = [ImagePickerUtils makeConfigurationFromOptions:options target:target];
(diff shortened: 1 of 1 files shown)


Johan-dutoit:
Barring some testing, looks okay.  I'll see where I can slice in some time to properly review and test, and will revert back if anything stands out.

Hannes-Endare:
https://developer.apple.com/documentation/photokit/phphotolibrary/3616113-presentlimitedlibrarypickerfromv
Prompts the user to update their limited library selection. It does not let the user pick any pictures.

lukebrandonfarrell:
@thibmaek this caused a crash:

Screenshot 2021-12-08 at 14 07 13

thibmaek:
I'm going to close this PR. We switched to another image picker library that handled this out of the box. This PR can be used as a reference for someone.

andrewzey:
@thibmaek Can you please elaborate? The original issue as stated (#1749) is still very much an open issue.

thibmaek:
@andrewzey yup that's true. This PR however was incomplete and partially incorrect so I closed it because it would not have fixed the issue. I have moved towards another library that does not have this issue so had no reason to keep maintaining this PR.

There is some stuff in this PR you could use to start your own though.

ccompton-ip:
@thibmaek This is also a deal breaker for me. Care to share which RN Library you've moved to that supports limited library on iOS?

Issue #1759, opened 2021-06-21
[🐛] Incompatible types: ReadableMap cannot be converted to WritableMap


goriverna:
Install library, compile with android studio, receive error from Android Studio on the file Utils.java:

error: incompatible types: ReadableMap cannot be converted to WritableMap
                assets.pushMap(getImageResponseMap(uri, options, context));

![image](https://user-images.githubusercontent.com/10176638/122719713-e418d400-d26e-11eb-9e74-aac1bf01a6d4.png)

### WORKAROUND

If you follow the suggestion made by android studio you can cast the method as "(WritableMAp)" 

`assets.pushMap((WritableMap) getImageResponseMap(uri, options, context));`

and

`assets.pushMap((WritableMap) getVideoResponseMap(uri, context));`

![image](https://user-images.githubusercontent.com/10176638/122720041-4245b700-d26f-11eb-8773-e020422d3938.png)

- Image Picker version: 4.0.3
- React Native version: 0.60.6
- Platform: Android
- Development Operating System: Mas OS Big Sur (Intel)
- Dev tools: Android Studio: 4.2.1, Compile SDK: 29, Gradle: 6.2, Android Gradle Plugin Version 3.5.3

ariva85:
Same issue here on RN 0.60.6

redocs:
Same problem.

RN 0.60.6

Carl0395:
I change the return of **getImageResponseMap** and **getVideoResponseMap** for WritableMap in node_modules/react-native-image-picker/android/src/main/java/com/imagepicker/Utils.java and fixed for now.

goriverna:
Yes @Carl0395 as my workaround, but we are waiting for a fix asap.

Jinxhao18:
+1

Johan-dutoit:
PR's welcome.

jackfiallos:
there's a PR that fix this issue here
https://github.com/react-native-image-picker/react-native-image-picker/pull/1841

could be good if the PR is merged since is affecting the latest version 4.3.0

Johan-dutoit:
Before merging that PR, I'd like to find out what's causing the issue.

This problem doesn't exist with the example app (on my side anyway), so it would be great if it the root cause can be identified and then we can make an informed decision as to if that PR is correct, or someone's setup is wrong.

jackfiallos:
[...]

PR #1815, opened 2021-09-07
Fix Incompatible types - Android (#1759)

rafirr:
## Motivation (required)
Change Incompatible types:  #1759 

## Test Plan (required)

```react-native run-android```


lukebrandonfarrell:
@rafirr I can merge this if you are able to share with me a scenario to reproduce? We are on RN 0.64 without issues

PR #1795, opened 2021-08-06
Added ability to include exif data for iOS and Android

lukebrandonfarrell:
Thanks for submitting a PR! Please read these instructions carefully:

- [X] Explain the **motivation** for making this change.
- [X] Provide a **test plan** demonstrating that the code is solid.
- [X] Match the **code formatting** of the rest of the codebase.
- [X] Target the `master` branch, NOT a "stable" branch.

## Motivation (required)

This adds the architecture for iOS and Android to optionally offer the ability to include exif data with the selected photos. The requires iOS library permissions, hence `includeExif` defaults to `false`.

## Test Plan (required)

These changes will be released to several thousand users in our production application via our fork.

PR #1816, opened 2021-09-09
iOS 14+ photos permission ask added 

lokesh020:
### Above iOS 14 version, Accessing photos permission is not asked from the user which should be the expected behaviour.

I have added functionality to ask permission from user for photos above iOS 14.
Please review ASAP as it was impacting users.

I am successfully done that and attaching screenshot for review purpose.

image0


ajayoncode:
I was also facing this issue and had to mange locally, Now this is really good solution 👍

ajayoncode (approved):

ajayoncode (approved):
Nice Work

lukebrandonfarrell:
@lokesh020 @ajayoncode Checkout:  https://github.com/react-native-image-picker/react-native-image-picker/pull/1795

lokesh020:
Hii @lukebrandonfarrell , pls suggest me the solution as above iOS 14 , photos permission is not asked by the library. If it's intended to do so, How can I achieve this as it is annoying. I have to change code locally and use it

Issue #1796, opened 2021-08-07
[🐛] LaunchImageLibrary returns {"didCancel": true} in response.


afanchamp:
### Description

LaunchImageLibrary returns User cancelled image picker in response after selecting the image from library. Facing this issue on Xiomi Note 9 and 10 and also on some Samsung phones. LaunchCamera is working fine but facing this issue when i select image from photo library.

JairajJangleSLIOT:
I am also getting response as `{"didCancel":true}`.

My device details:
Xiaomi Redmi K20
Android 11
MIUI Global 12.1.4

elliotclements-mendix:
Same issue here Android 12

PR #1854, opened 2021-10-22
fix(android): LaunchImageLibrary returns {"didCancel": true} in response

liaojh554:

## Motivation (required)

fix(android): LaunchImageLibrary returns {"didCancel": true} in response. #1796

## Test Plan (required)

react-native run-android

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -152,7 +152,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule implements Act
         }
 
         try {
-            currentActivity.startActivityForResult(Intent.createChooser(libraryIntent, null), requestCode);
+            currentActivity.startActivityForResult(libraryIntent, requestCode);
         } catch (ActivityNotFoundException e) {
             callback.invoke(getErrorMap(errOthers, e.getMessage()));
             this.callback = null;


iamadityaaz:
I tried the same, but It doesn't works for me.
@liaojh554

afizboltz:
maybe you need to rebuild the app again

iamadityaaz:
@afizboltz Done it but still no  success.
Also added
```
    <queries>
        <intent>
            <action android:name="android.media.action.IMAGE_CAPTURE" />
        </intent>
        <intent>
            <action android:name="android.intent.action.GET_CONTENT" />
            <data android:mimeType="image/*" />
        </intent>
        <intent>
            <action android:name="android.intent.action.PICK" />
            <data android:mimeType="image/*" />
        </intent>
        <intent>
            <action android:name="android.intent.action.CHOOSER" />
        </intent>
    </queries>
```
Following https://developer.android.com/guide/topics/manifest/queries-element and https://github.com/react-native-image-picker/react-native-image-picker/issues/1393#issuecomment-713299590

iamadityaaz:
Updated 
ImagePickerModule.java
Line : 155 
            currentActivity.startActivityForResult(Intent.createChooser(libraryIntent, null), requestCode);
to 
            currentActivity.startActivityForResult(libraryIntent, requestCode);


> Output:

It hangs (Takes more than 2 minutes) while showing Options to choose Gallery apps  (Photos, Gallery, etc) and when it open after selecting the image, App crashes.

afizboltz:
> Updated ImagePickerModule.java ` currentActivity.startActivityForResult(Intent.createChooser(libraryIntent, null), requestCode);` to
> 
> ` currentActivity.startActivityForResult(libraryIntent, requestCode);`
> 
> > Output:
> 
> It hangs (Takes more than 2 minutes) while showing Options to choose Gallery apps (Photos, Gallery, etc) and when it open after selecting the image, App crashes.

from my case, redmi device no delay showing the gallery

JairajJangleSLIOT:
Can we get an estimate of when this PR will be merged?

 It would be good if we get some estimation to allow us to make the decision of sticking with this module or selecting another module instead.

lukebrandonfarrell:
@JairajJangleSLIOT I can only merge this if @liaojh554 or @afizboltz shares a reliable scenario to replicate this issue

afilp:
I have the same problem (Xiaomi 9T Pro, Android 11), it always returns this object:
{
    "didCancel": true
}

But I am not sure how to help, we really need this to be fixed. 
Thanks!

afilp:
And yes, I confirm that the PR solves the problem! I changed this one line of code and rebuilt/rerun the project and now I can add images without problems.

If there are no other side-effects it would be nice to merge this, thanks a lot to all!

JairajJangleSLIOT:
After applying the changes as done in this PR's commit, I am still getting response as `{"didCancel":true}`. Tried clearing yarn and Gradle cache and reapplied the changes to the `ImagePickerModule.java`, still, the same response. So, in my experience, this PR did not fix the issue at least on my device. 

My device details:
Xiaomi Redmi K20, 
Android 11
MIUI Global 12.1.4

afilp:
Hello, do you believe we could approve this? Thanks a lot!

lukebrandonfarrell:
Could you walk me through how this fixes the issue? I'm unaware of what `Intent.createChooser` is doing?

lukebrandonfarrell:
Also from @JairajJangleSLIOT comment, seems like this isn't fixing the issue:

```
After applying the changes as done in this PR's commit, I am still getting response as {"didCancel":true}. Tried clearing yarn and Gradle cache and reapplied the changes to the ImagePickerModule.java and rebuilt the project, still, the same response. So, in my experience, this PR did not fix the issue at least on my device.
```

I'm worried we aren't addressing the core of the issue. Do you know why it returns `{"didCancel":true}` on Xiaomi devices?

JairajJangleSLIOT:
@lukebrandonfarrell @afilp **Turning on MIUI optimization** on Xiaomi devices in Developer Options fixes this issue.

But the side effect of doing so is that USB debugging gets disabled. So, I believe that a normal Xiaomi device where the user hasn't cared to fiddle with this option won't have this issue.

I have tested this workaround of toggling **MIUI Optimization** on _Redmi K20_ and _Redmi Note 10 Pro_, which works on both devices. Turning it on does fix this issue and so Turning it off creates this issue.

Xiaomi has made it very tricky to access this **MIUI optimization** option. 
To access this **MIUI optimization** option:
1. Go to developer options
2. Scroll to the bottom-most of this screen
3. Tap on **Reset to default values** multiple times (5-6 times)
4. **Turn on MIUI optimization** option will appear below
5. If you are using USB Debugging, you have to disable **Turn on MIUI optimization** which causes this issue.

@lukebrandonfarrell I believe this may be the core reason for this issue. 

I hope this information helps.

afilp:
@JairajJangleSLIOT
 
 Excellent finding and explanation!
 
 I concur, everything works OK by turning ON this optimization.
 
 If there is a way to get this option in native code and do things differently, that would be great.
 
 @lukebrandonfarrell Since the PR proposal did work in my Xiaomi 9T Pro, even with MIUI optimization OFF, what is the risk of accepting it? Does it cause other problems to other devices? If not, why not apply it and have more supported Xiaomi devices this way? Maybe there is some other option also in @JairajJangleSLIOT 's device that cause the problem combined. I agree we need to find that too, but we may be better off to release this PR till then, to support more devices. Thanks!

lukebrandonfarrell:
@afilp I'm a little nervous about merging this for a few reasons:

- I'm not sure what `Intent.createChooser` is doing, maybe the author could write a more detailed description for their PR? @liaojh554 
- I don't know if this change may impact and cause issues on other Android devices.
- There has been comments suggesting that this doesn't fix the issue: https://github.com/react-native-image-picker/react-native-image-picker/pull/1854#issuecomment-994292445
- Finally, I don't think this addresses the core of the issue. Once we have an explication of why `MIUI Optimization` and which part of our code is causing this issue in detail, then I may consider merging and testing this, but at the point the author has left the PR description and details pretty sparse.

liaojh554:
@lukebrandonfarrell  hi, I think the cause of this problem:
1. Intent.createChooser
   CreateChooser () calls new Intent(ACTION_CHOOSER), which returns an Intent object and pops up as an Activity of ChooserActivity.
2. Android assumes that activities between tasks cannot pass data
3. MIUI Optimization
   I think MIUI Optimization has changed ChooserActivity, possibly setting FLAG_ACTIVITY_NEW_TASK so that the caller of startActivityForResult() is not in the same task as the called, Throw RESULT_CANCELED

​

lukebrandonfarrell:
@liaojh554 Makes sense, let me test this to make sure there is no unintended behaviour, then happy to merge

afilp:
@lukebrandonfarrell If you need any help with this (testing) please let me know, thank you!

lukebrandonfarrell:
Seems like this is quite a popular fix, going to merge https://github.com/react-native-image-picker/react-native-image-picker/pull/1928 instead (which does the same thing)

Issue #1699, opened 2021-05-11
[🐛] NullPointerException: android.net.Uri.getLastPathSegment(), from getResponseMap(Utils.java:339)


thanakij:
### Description

Crashlytics has reported an issue related to NullPointerException on **Utils.java (line 339)**

`String fileName = uri.getLastPathSegment();`

```
Fatal Exception: java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=13002, result=-1, data=Intent { dat=content://com.sec.android.gallery3d.provider/union/image/item/15652 flg=0x1 (has extras) }} to activity {com.x/com.x.MainActivity}:
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.net.Uri.getLastPathSegment()' on a null object reference
```

### How to repeat issue and example

I didn't find this error myself. Testing on my local Android devices always work fine but some portion of our users found errors according to Crashlytics report.

### Solution

Crashlytics revealed that the crash is from devices running **Android 8.1.0**. I have gone through the code and am convinced that there is an exception somewhere in `resizeImage()` as this method has a big try-catch wrapped around it. So it would return null and cause the crash in the next following line in `getResponseMap()`.

### Additional Information

- Image Picker version: 3.5.0
- React Native version: 0.62.2
- Platform: Android
- Development Operating System: MacOS
- Dev tools: N/A

SumantaBee:
Having same issue
`{"errorCode": "others", "errorMessage": "Attempt to invoke virtual method 'java.lang.String android.net.Uri.getLastPathSegment()' on a null object reference"}
`

PR #1862, opened 2021-10-27
fix(android): return the original 'uri' instead of null when resizeImage has failed

thanakij:

- [  ] Provide a **test plan** demonstrating that the code is solid.

## Motivation (required)

From #1699, I believe it is not safe to close the issue as latest code the report has already taken into `NullPointerException` with a more broad `RuntimeException`. Anyway, I find that it can be more stronger by instead of returning null (in `resizeImage` method which would trigger NPE in the first place), we can simply return the original uri.

## Test Plan (required)

I am not really sure if I can have a concrete test plan for this one-line change. It is a simple improvement to the existing codebase.

### DIFF
--- a/android/src/main/java/com/imagepicker/Utils.java
+++ b/android/src/main/java/com/imagepicker/Utils.java
@@ -212,7 +212,7 @@ public class Utils {
 
         } catch (Exception e) {
             e.printStackTrace();
-            return  null;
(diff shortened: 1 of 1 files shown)


lukebrandonfarrell:
This makes a lot of sense, thanks for the contribution!

lukebrandonfarrell (approved):

PR #1877, opened 2021-11-26
fix: selectionLimit assets return on android

imdbsd:

## Motivation (required)

Hai, im using the `selectionLimit` options on to limit the maximum selection file that user can do. On ios, this work fine as video below.

https://user-images.githubusercontent.com/16449020/143539168-ba2e0f49-4805-47f9-80fb-047c9a85bd72.mp4

it block the user from select more than the limit and return the selection as the limit. But on android, user can select as much as possible and the return is all the user selection.

https://user-images.githubusercontent.com/16449020/143539308-57e1b796-a018-4453-bad4-e585ed896fa0.mp4

as what i read from SO https://stackoverflow.com/questions/33604951/maximum-image-selection-limit-from-gallery-android, it need to validate again the selection file as the limit. In this PR, im improve to return only the amount of selected image as set in selectionLimit options. Please see the video below.

https://user-images.githubusercontent.com/16449020/143539552-04f9907c-8fb0-43c0-a8fd-b8b2bda32d18.mp4

## Test Plan (required)

- confused on how to add test

### DIFF
--- a/android/src/main/java/com/imagepicker/Utils.java
+++ b/android/src/main/java/com/imagepicker/Utils.java
@@ -436,8 +436,15 @@ public class Utils {
 
     static ReadableMap getResponseMap(List<Uri> fileUris, Options options, Context context) throws RuntimeException {
         WritableArray assets = Arguments.createArray();
+        int size;
+        if(fileUris.size() > options.selectionLimit) {
+            size = options.selectionLimit;
+        }
+        else {
+            size = fileUris.size();
+        }
 
-        for(int i = 0; i < fileUris.size(); ++i) {
+        for(int i = 0; i < size; ++i) {
             Uri uri = fileUris.get(i);
 
             if (isImageType(uri, context)) {


Johan-dutoit:
Hey, thanks for the detailed PR.

I have initial concerns about this, but more than happy to discuss.

- This PR introduces a soft limit (i.e. a user will still be able to select more than the limit hit)
- The code/app will only get the limit, can we guarantee order of images, etc?
- User will see something different to what they selected (causing confusion)

Thoughts?

imdbsd:
Hi, thanks for the feedbck @Johan-dutoit 

- yes, this pr will act as soft limit from the user selection. 

- For order of the image, it should be same, since im only limit the returned size without doing any sort. 

- For point number 3, im agree that this will lead to bad ux for the user, im looking for other native android library, and they implement their own limit on the picker. Im happy to try implement it but will take some times, since im not fluent  enough with android and java, so this is the current help that i can do.

lukebrandonfarrell:
I think this is out of scope, the Android library doesn't let you limit the number of picker images, you should do `results.slice` when you get the result back from the picker.

PR #1881, opened 2021-12-04
fix(android): Add filename extension to `fileName` for videos from library

chrisbobbe:

## Motivation (required)

fix(android): Add filename extension to `fileName` for videos from library

When uploading videos from the library, we use a `content://` URL,
and we've been using `.getLastPathSegment()` on that to create
`fileName`.

In my experience, that `fileName` hasn't contained a filename
extension. So, when that's the case, add one, based on the MIME
type.


lukebrandonfarrell:
@Johan-dutoit this is technically a breaking change no? Maybe we want to add a new property i.e. `"extension"` and include it in there instead.

lukebrandonfarrell:
@chrisbobbe What is the use case here, why do you need the file name extension in your app?

Johan-dutoit:
In it's current state, yes, it would be a breaking change.

lukebrandonfarrell:
@chrisbobbe are you open to expanding on this to return `extension` as a separate key?

chrisbobbe:
> @chrisbobbe are you open to expanding on this to return `extension` as a separate key?

Happy to make a less breaking change, yes! 🙂 For example, yes, things could break for an app that's already adding an extension to `fileName` without checking to see if it has one already. No one wants two extensions.

Sure, a new, separate key sounds good. What do you think of the key being a full filename including extension, instead of the extension by itself? Then callers don't have to implement their own logic to put `fileName` and the extension together. We can name it something like `cookedFileName`, and maybe eventually it can replace the current `fileName`. I think it's much more natural to want a filename with an extension than without. And there's nothing magical about the current value at `fileName`; it doesn't uniquely identify a file, like the URL does.

If we're adding that separate key, I think it might be kind of awkward to document and use if it only treats the videos-on-Android case. (What about photos on iOS?) It'd be nice if the key gave a meaningful value on all types of responses. I'm thinking we could add it in the thin JavaScript wrapper instead of at the native layer, to accomplish that.

The logic would be pretty simple: add an extension if `fileName` doesn't have one already, and if we can find an appropriate one; otherwise, just copy `fileName`. We can use `getExtension` from [`mime`](https://www.npmjs.com/package/mime), maybe the "Lite Version" (see the doc) that's reportedly ~2.5KB instead of 8KB.

Then we could document the new key with something like "Like `fileName`, but after trying to add an appropriate extension, to make it a better suggestion for a filename."

lukebrandonfarrell:
@chrisbobbe Since we already have MIME type returned we could do this all on the JavaScript layer? We just add `fileName` + the end of the extension `mime?.split("/")[1]` and return a new property called `fileNameWithExtension` ? Then there is no need to add anymore libraries, or to modify any native code. Would that work?

chrisbobbe:
TL;DR: I think the solution I've proposed [above](https://github.com/react-native-image-picker/react-native-image-picker/pull/1881#issuecomment-991418749) has all the qualities we'd want. 🙂

-----

> we could do this all on the JavaScript layer?

That's right. The solution I've suggested has this property. Please see:

> I'm thinking we could add [the new key] in the thin JavaScript wrapper instead of at the native layer

The [`mime`](https://www.npmjs.com/package/mime#user-content-lite-version) NPM package I suggested exposes JavaScript functions we would call. It looks like the one we'd use is `getExtension`.

You asked if the following would work:

> We just add `fileName` + the end of the extension `mime?.split("/")[1]`

It wouldn't. The part that comes after "/" in the MIME type is called the subtype; see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#structure_of_a_mime_type) and the [spec](https://datatracker.ietf.org/doc/html/rfc6838). The subtype isn't a valid filename extension in general. In fact, filename extensions can't in general be extracted directly from MIME-type strings.

The suggested transformation would give bad results:

| MIME type  | Proposed name (bad) | Better name |
| --- | --- | --- |
| `text/plain` | "foo.plain" | "foo.txt" |
| `image/svg+xml` | "foo.svg+xml" | "foo.svg" (see [registration document](https://www.iana.org/assignments/media-types/image/svg+xml)) |
| `video/iso.segment` | "foo.iso.segment" | "foo.m4s" (see [registration document](https://www.iana.org/assignments/media-types/video/iso.segment)) |

We need a data source for mapping MIME types to extensions in common use. It won't be practical to [hard-code something ourselves](https://github.com/react-native-image-picker/react-native-image-picker/blob/v4.6.0/android/src/main/java/com/imagepicker/Utils.java#L288-L298) that gives correct results for all MIME types we expect to see.

The set of valid MIME types and common extensions change over time, and that's another reason to use a library, as long as it stays maintained and reasonably up-to-date. Again, I think the ["Lite Version"](https://www.npmjs.com/package/mime#user-content-lite-version) part of the `mime` NPM package would probably work well. And that package doesn't bring in any additional dependencies of its own.

lukebrandonfarrell:
@chrisbobbe Gotcha, starting to follow now! I think this sounds good, @Johan-dutoit what do you think of this?

Johan-dutoit:
This is an interesting problem, which is pushing me towards this being out of scope (although, happy to discuss).

If it's in the JS land, then the users can do it themselves.  This allows this package to stay "lean" and provides users with an opt-in solution.

With that being said, I feel like our goalposts should look something like this:
- Ensure that we return the correct data for a user to calculate the mimetype (using the 3rd party lib above)
- Provide examples on how to do it/update docs 
- Remove our own half baked solution

chrisbobbe:
Thanks for your thoughts on this! 🙂 I've just sent #1895 to demonstrate an implementation; I'm hoping that can help move the discussion forward.

jfries289:
There are 8 possible return cases between iOS or Android, Video or Image, and gallery or capture. In 7 of those case, the `fileName` returned is a fully valid file name with extension and the `uri` is a full, valid file path, and in one of those case, Android->Gallery->Video, it's a content uri and a file name without an extension.  This is a bug and not a feature. It is not good design to have an exception in your return object in 1 out of 8 cases. 

For the sanity of developers who do and will make use of your very useful library, please consider changing that.

PR #1889, opened 2021-12-09
feat(asset): return id of asset for iOS and Android

helenaford:

## Motivation (required)

ID can be used to detect if the user has selected the image or video again, useful when app supports multiple images or videos to be uploaded

## Test Plan (required)

Android:

Screenshot 2021-12-09 at 15 59 38
Screenshot 2021-12-09 at 15 57 57
Screenshot 2021-12-09 at 15 57 18

iOS:

### DIFF
--- a/android/src/main/java/com/imagepicker/Utils.java
+++ b/android/src/main/java/com/imagepicker/Utils.java
@@ -421,8 +421,9 @@ public class Utils {
 
         if(options.includeExtra) {
           String datetime = getDateTimeExif(uri, context);
-          // Add more exif data here ...
+          // Add more extra data here ...
           map.putString("timestamp", datetime);
+          map.putString("id", fileName);
         }
 
         return map;
@@ -478,7 +479,7 @@ public class Utils {
       }
     }
 
-    static ReadableMap getVideoResponseMap(Uri uri, Context context) {
+    static ReadableMap getVideoResponseMap(Uri uri, Options options, Context context) {
         String fileName = uri.getLastPathSegment();
         WritableMap map = Arguments.createMap();
         map.putString("uri", uri.toString());
@@ -486,6 +487,11 @@ public class Utils {
         map.putInt("duration", getDuration(uri, context));
         map.putString("fileName", fileName);
         map.putString("type", getMimeType(uri, context));
+
+        if(options.includeExtra) {
+          map.putString("id", fileName);
+        }
+
         return map;
     }
 
@@ -502,7 +508,7 @@ public class Utils {
                 uri = resizeImage(uri, context, options);
                 assets.pushMap(getImageResponseMap(uri, options, context));
             } else if (isVideoType(uri, context)) {
-                assets.pushMap(getVideoResponseMap(uri, context));
+                assets.pushMap(getVideoResponseMap(uri, options, context));
             } else {
                 throw new RuntimeException("Unsupported file type");
             }
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -164,13 +164,14 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
         NSString *creationDate = [formatter stringFromDate:phAsset.creationDate];
         
         asset[@"timestamp"] = creationDate;
-        // Add more exif data here ...
+        asset[@"id"] = phAsset.localIdentifier;
+        // Add more extra data here ...
     }
     
     return asset;
 }
 
--(NSMutableDictionary *)mapVideoToAsset:(NSURL *)url error:(NSError **)error {
+-(NSMutableDictionary *)mapVideoToAsset:(NSURL *)url phAsset:(PHAsset * _Nullable)phAsset error:(NSError **)error {
     NSString *fileName = [url lastPathComponent];
     NSString *path = [[NSTemporaryDirectory() stringByStandardizingPath] stringByAppendingPathComponent:fileName];
     NSURL *videoDestinationURL = [NSURL fileURLWithPath:path];
@@ -201,13 +202,19 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
           }
         }
     }
-    
+
     NSMutableDictionary *asset = [[NSMutableDictionary alloc] init];
     asset[@"fileName"] = fileName;
     asset[@"duration"] = [NSNumber numberWithDouble:CMTimeGetSeconds([AVAsset assetWithURL:videoDestinationURL].duration)];
     asset[@"uri"] = videoDestinationURL.absoluteString;
     asset[@"type"] = [ImagePickerUtils getFileTypeFromUrl:videoDestinationURL];
     asset[@"fileSize"] = [ImagePickerUtils getFileSizeFromUrl:videoDestinationURL];
+
+    if (phAsset) {
+      asset[@"id"] = phAsset.localIdentifier;
+      // Add more extra data here ...
+    }
+
     
     return asset;
 }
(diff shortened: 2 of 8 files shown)


ios/ImagePickerManager.m
```diff
@@ -331,32 +338,25 @@ - (void)imagePickerController:(UIImagePickerController *)picker didFinishPicking
 {
     dispatch_block_t dismissCompletionBlock = ^{
         NSMutableArray<NSDictionary *> *assets = [[NSMutableArray alloc] initWithCapacity:1];
+        PHAsset *asset = nil;
+
+        // If include extra, we fetch the PHAsset, this required library permissions
+        if([self.options[@"includeExtra"] boolValue]) {
+          asset = [ImagePickerUtils fetchPHAsset:info];
```
Johan-dutoit, line 345: Can we catch an error and report it if permissions haven't been granted?

helenaford: ah, good shout! could we do this as a follow-up p/r as this logic of fetching the asset has already been merged/released?

android/src/main/java/com/imagepicker/Utils.java
```diff
@@ -478,14 +479,19 @@ static ReadableMap getImageResponseMap(Uri uri, Options options, Context context
...
         String fileName = uri.getLastPathSegment();
         WritableMap map = Arguments.createMap();
         map.putString("uri", uri.toString());
         map.putDouble("fileSize", getFileSize(uri, context));
         map.putInt("duration", getDuration(uri, context));
         map.putString("fileName", fileName);
         map.putString("type", getMimeType(uri, context));
+
+        if(options.includeExtra) {
+          map.putString("id", fileName);
```
lukebrandonfarrell, line 492: So the `includeExtra` flag is specifically for values which need permissions, does `id` need permissions to be accessed on iOS? I'm not sure if it's necessary for the `options.includeExtra` check here. Thoughts?

ios/ImagePickerManager.m
```diff
@@ -164,13 +164,14 @@ -(NSMutableDictionary *)mapImageToAsset:(UIImage *)image data:(NSData *)data phA
         NSString *creationDate = [formatter stringFromDate:phAsset.creationDate];
         
         asset[@"timestamp"] = creationDate;
-        // Add more exif data here ...
+        asset[@"id"] = phAsset.localIdentifier;
+        // Add more extra data here ...
```
lukebrandonfarrell, line 168: Same comment as above. Does this need permissions to be accessed? I'm guessing as it is using PHAsset it does, right?

ios/ImagePickerManager.m
```diff
@@ -201,13 +202,19 @@ -(NSMutableDictionary *)mapVideoToAsset:(NSURL *)url error:(NSError **)error {
...
     asset[@"fileName"] = fileName;
     asset[@"duration"] = [NSNumber numberWithDouble:CMTimeGetSeconds([AVAsset assetWithURL:videoDestinationURL].duration)];
     asset[@"uri"] = videoDestinationURL.absoluteString;
     asset[@"type"] = [ImagePickerUtils getFileTypeFromUrl:videoDestinationURL];
     asset[@"fileSize"] = [ImagePickerUtils getFileSizeFromUrl:videoDestinationURL];
+
+    if (phAsset) {
+      asset[@"id"] = phAsset.localIdentifier;
+      // Add more extra data here ...
+    }
```
lukebrandonfarrell, line 216: Tomorrow I'll be adding video exif data here, as we need it in our product, this is a great update! But I wonder in the future if we could use some abstraction for `mapVideoToAsset` and `mapImageToAsset`, not suggesting a change here, just want to hear your thoughts

ios/ImagePickerManager.m
lukebrandonfarrell: @Johan-dutoit I'm pretty sure the library returns an error further up the chain if permissions are NOT granted but you passed `includeExtra`, but I agreed with @helenaford this should be done in a separate PR.

Maybe we could enable project boards on this repo @Johan-dutoit and start adding some improvement ideas there?

ios/ImagePickerUtils.m
```diff
@@ -156,4 +156,18 @@ + (UIImage*)resizeImage:(UIImage*)image maxWidth:(float)maxWidth maxHeight:(floa
     return newImage;
 }
 
++ (PHAsset *)fetchPHAsset:(NSDictionary<NSString *,id> *)info
```
lukebrandonfarrell, line 159: I may be wrong, but is this method only for iOS 13 and below? As iOS 14 uses the PHPhotoPicker callback, if so shall we rename this to something like `fetchPHAssetFromAssetURL` or `fetchPHAssetOnOS13` ? Lmk!

lukebrandonfarrell (commented):
Some questions, but looks good to me!

android/src/main/java/com/imagepicker/Utils.java
helenaford: @lukebrandonfarrell yeah, no permission is needed for Android. I was just trying to be consistent with iOS? Does fetching the timestamp on Android require permissions?

Can easily remove it, and add a note on the readme to say flag is only needed for iOS

ios/ImagePickerManager.m
helenaford: yeah, needs permissions due to the PHAsset on iOS

ios/ImagePickerManager.m
helenaford: yeah I think that could be good!! Will reduce duplication and make it clearer which props are shared across videos and images

ios/ImagePickerManager.m
Johan-dutoit: Agree on both points.

android/src/main/java/com/imagepicker/Utils.java
lukebrandonfarrell: @helenaford fetching the timestamp doesn't require permission on Android, it's a hard decision because consistency simplifies the understanding of how the API works to consumers, but we could technically remove it on Android. I say let's keep it for now and I can remove it in my next PR for Android, as I will be adding `datetime` to video next week.

ios/ImagePickerManager.m
lukebrandonfarrell: Makes sense, all is good here 🙌

ios/ImagePickerUtils.m
lukebrandonfarrell: Sounds good! Let's update that, then I'll merge 🎉 Thanks for the contributions !

helenaford: @lukebrandonfarrell updated 🙌

lukebrandonfarrell (approved):
LGTM!

helenaford:
@Johan-dutoit @lukebrandonfarrell thanks both! Will be able to upgrade now to the latest library version 😎

Seanmclem:
> @Johan-dutoit @lukebrandonfarrell thanks both! Will be able to upgrade now to the latest library version 😎

Thanks for this, but I'm running 4.6.0, and I don't get an `id` back. Might there be anything I missed? I'm using `launchImageLibrary` data. Maybe because I'm on simulator?

I get 
```
{
   "fileName":"DE029C1F-51F5-48D7-9129-3F951FD3A515.jpg",
   "fileSize":845465,
   "height":667,
   "type":"image/jpg",
   "uri":"file:///Users/seanclements/Library/Developer/CoreSimulator/Devices/E50F3981-2254-427A-950D-3022C5583F08/data/Containers/Data/Application/D6CC86C7-EC03-485C-A9B8-DDD321145FB8/tmp/DE029C1F-51F5-48D7-9129-3F951FD3A515.jpg",
   "width":1000
}
```

lukebrandonfarrell:
@Seanmclem you need to pass the `includeExtra` flag

Issue #1856, opened 2021-10-24
[🐛] Returns invalid mime types (`image/jpg`) instead of the correct `image/jpeg`


dylanpyle:
### Description

When you select a photo from the library or camera, `response.assets` includes an array of assets and metadata e.g:

```
[
    {
        "type": "image/jpg",
        "fileName": "8F471139-AB15-457B-A817-EF41BF621231.jpg",
        "width": 3024,
        "height": 4032,
        "uri": "file:///var/mobile/Containers/Data/Application/CA9B8D34-12F7-41A6-9364-FEBF5C536BAD/tmp/8F471139-AB15-457B-A817-EF41BF621231.jpg",
        "fileSize": 6718368
    }
]
```

Though the readme doesn't use the phrase "mime type" or promise to adhere to any specific standard, the format of the `type` field suggests it's supposed to be a "mime type", which is a specific [set of standards defined by IANA/IETF](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)

`image/jpg` isn't a valid mime type (`image/jpeg` is the correct type to use), and most databases (e.g. [mime-db](https://github.com/jshttp/mime-db/blob/master/db.json#L7039)) can't decode `image/jpg` into something useful.

This may the case for other types (I haven't checked) but is definitely wrong for JPGs.

### How to repeat issue and example

Select an image and inspect the `response.assets` array.

### Solution

Return the valid `"type": "image/jpeg"` for JPG files.

### Additional Information

- Image Picker version: 4.1.1
- React Native version: 0.63.3
- Platform: iOS
- Development Operating System: macOS
- Dev tools: Xcode 12.5.1

yberstad:
I fixed this issue by creating a patch-file, `react-native-image-picker+4.3.0.patch:`
```
diff --git a/node_modules/react-native-image-picker/ios/ImagePickerUtils.m b/node_modules/react-native-image-picker/ios/ImagePickerUtils.m
index a611b74..3ae322a 100644
--- a/node_modules/react-native-image-picker/ios/ImagePickerUtils.m
+++ b/node_modules/react-native-image-picker/ios/ImagePickerUtils.m
@@ -93,13 +93,13 @@ + (NSString*) getFileType:(NSData *)imageData
     [imageData getBytes:&firstByte length:1];
     switch (firstByte) {
       case firstByteJpg:
-        return @"jpg";
+        return @"jpeg";
[...]

PR #1890, opened 2021-12-10
fix (iOS): correct MIME-type for JPEG files

AleksandrTermenzhy:

## Motivation (required)

Correct MIME-type for [JPEG files](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image) is 'image/jpeg' while file extensions vary (jpeg, jpg etc.).

It may lead to some edge cases with correct decoding like the one described [here](https://github.com/react-native-image-picker/react-native-image-picker/issues/1856)

Only iOS version seems to be affected.

## Test Plan (required)
Without fix
Without fix

With fix
With fix

P. S. I'll appreciate if someone more familiar with iOS code suggest a more elegant solution or find potential pitfalls here, 'cause I'm new to it.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -133,7 +133,11 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
     }
     
     NSMutableDictionary *asset = [[NSMutableDictionary alloc] init];
-    asset[@"type"] = [@"image/" stringByAppendingString:fileType];
+    if ([fileType isEqualToString:@"jpg"]) {
+        asset[@"type"] = [@"image/" stringByAppendingString:@"jpeg"];
+    } else {
+        asset[@"type"] = [@"image/" stringByAppendingString:fileType];
+    }
 
     NSString *fileName = [self getImageFileName:fileType];
     NSString *path = [[NSTemporaryDirectory() stringByStandardizingPath] stringByAppendingPathComponent:fileName];


lukebrandonfarrell:
@AleksandrTermenzhy the issue here is that the file type is JPG and it is returning jpg? Seems like correct behaviour to me

AleksandrTermenzhy:
@lukebrandonfarrell currently field `type` is set to `image/jpg` for jpg/jpeg files while the standard MIME-type is `image/jpeg`.
And this is not a big deal when you use it only within your app. But it could be a problem when dealing with some external services which expect MIME-types to be a standard ones.

lukebrandonfarrell:
Could you link me to this standard? I'd be in favour of returning what we get from the system, as arbitrary replacement could cause confusion.

AleksandrTermenzhy:
@lukebrandonfarrell Actually I posted link in pull request description. [Here](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image) you go.

As far as I understood from code, there is no "type we get from the system". It's defined based on the first byte of the file in `getFileType` util.
But again, please correct me if I'm wrong and I'll appreciate any help with this as I'm not well familiar with Obj-C stuff.

lukebrandonfarrell:
@AleksandrTermenzhy gotcha, this is starting to make more sense now, apologies for the confusion! I think it would be better if we address the root of the problem here, and change the typing in [getFileType](https://github.com/react-native-image-picker/react-native-image-picker/blob/82fe096ad159d0e30ce6c524ce21573a9cfcf5f3/ios/ImagePickerUtils.m#L86) to return a `jpeg` although there are a few other places in the app where 'jpg' string is used for comparison, so those will need updated [etc.](https://github.com/react-native-image-picker/react-native-image-picker/blob/main/ios/ImagePickerManager.m#L129)

Looking into this further I am wondering why we even have `getFileType` in the first place and why we don't just use [getFileTypeFromUrl](https://github.com/react-native-image-picker/react-native-image-picker/blob/82fe096ad159d0e30ce6c524ce21573a9cfcf5f3/ios/ImagePickerUtils.m#L106) to allow iOS to determine the type.

Aure77:
Use this patch to fix `getFileType` return "jpeg":
https://github.com/react-native-image-picker/react-native-image-picker/issues/1856#issuecomment-977203119

lukebrandonfarrell:
Could we make a PR to change `jpg` to `jpeg`? and Also I think that patch alone would break other things, as we have some `if jpg` conditions in the code.

AleksandrTermenzhy:
@Aure77 @lukebrandonfarrell 
We surely can make fix to return `jpeg` instead of `jpg` from `getFileType`. But this change will also affect file extension, not only MIME-type - it will become `someFileName.jpeg` instead of `someFileName.jpg`. There is a non-zero possibility that someone relies on the extension as well.

But If this is OK for us, I can change PR that way.

jfries289:
As noted in the linked documentation, the `jpeg` in standard mime type `image/jpeg` does not refer specifically to the `.jpeg` extension, but is used for compressed photo data types that include the extensions `.jpg`, `.jpeg`, `.jpe`, `.jif`, `.jfif`.  Please see https://stackoverflow.com/a/37266399/2615373. Sending the incorrect mime type can (and has for us) caused issues with other libraries, browsers, and servers.

lukebrandonfarrell:
I agreed with this @jfries289 although patching `asset[@"type"]` is not the solution here. We would need a PR which done something like: https://github.com/react-native-image-picker/react-native-image-picker/pull/1890#issuecomment-992582967

AleksandrTermenzhy:
@lukebrandonfarrell Please advise how to handle the following:
... this change will affect not only MIME-type, but file extension as well - it will become someFileName.**jpeg** instead of someFileName.**jpg**. There is a non-zero possibility that someone relies on the extension as well.

Or we can just ignore that?

PR #1895, opened 2021-12-16
Add `cookedFileName`, to give `fileName` with extension when possible

chrisbobbe:
An alternative to #1881, described at https://github.com/react-native-image-picker/react-native-image-picker/pull/1881#issuecomment-991418749.

I understand from https://github.com/react-native-image-picker/react-native-image-picker/pull/1881#issuecomment-992362255 that we're thinking of leaving out this fix, to keep the library lean. I wanted to offer this implementation, to show that it doesn't need a very large amount of code or complexity. If you're interested in looking over the proposed code/docs changes, I'd also appreciate some eyes on the commit message. Thanks! 🙂


lukebrandonfarrell:
I'll give this a review @chrisbobbe !

chrisbobbe:
> I'll give this a review @chrisbobbe !

Thanks, Luke! 🙂

Issue #1532, opened 2020-12-13
[🐛] illegal callback invocation from navite module when tap twice by photo on iOS


alexeyvax:
### Description

Hi there!
Thank you for this awesome library, but I found a small issue. 

When the user open gallery with photos by `launchImageLibrary` method and tap twice by any photo, the red screen with error will appear.

The full error message is: `illegal callback invocation from navite module. This callback type only permits a single invocation from native code.`

I have a custom button, there is a code below: 
```
const DEFAULT_OPTIONS = {
  quality: 1,
  cameraType: 'back',
  mediaType: 'photo',
  allowsEditing: false,
  storageOptions: {
    skipBackup: true,
    path: 'images',
    waitUntilSaved: true,
    cameraRoll: false,
  },
  noData: true,
};
const someImagePickerCallback = async () => {
  ... some code here
};

<Button
  onPress={() => {
    ImagePicker.launchImageLibrary(DEFAULT_OPTIONS, someImagePickerCallback)
  }}
/>
```

### How to repeat issue and example

- Create a Button with onPress handler
- inside onPress should call `ImagePicker.launchImageLibrary` (like code above)
- gallery with photos will appear
- tap twice to any photos inside this gallery
- see red screen with error, which I described above

![Снимок экрана 2020-12-13 в 8 08 19 AM](https://user-images.githubusercontent.com/22244018/102004620-22ab6000-3d1b-11eb-9fd6-05b644ab0618.png)

### Additional Information

- Image Picker version: ^2.3.1
- React Native version: ^0.63.3
- Platform: iOS
- Development Operating System: MacOS
- Dev tools: Xcode Version 12.2

If you need more info please let me know.
Thank you!

z4gunn:
Able to reproduce on the following:

React Native: v0.61.0
Image Picker: v3.1.3
iOS: v14.2

Hadi-G:
Same issue

Hadi-G:
Same issue

SimajjiHans:
Any update here?

NitinTayal007:
getting the same issues, need some solution urgently...

rdjuric:
Same issue 💀

ozberkctn:
Same issue

AxeldeArias:
Same issue

penghouho:
Can reproduce in

react-native-image-picker@3.6.0
react-native@0.63.4

Desintegrator:
same

[...]

PR #1922, opened 2022-01-18
fix ios illegal callback invocation from navite module

Desintegrator:
fix "illegal callback invocation from navite module" when launchImageLibrary and user tap twice on photo before PickerViewController close

## Motivation (required)

this simple change should fix ios "illegal callback invocation from navite module" error like describe in some issues like [this](https://github.com/react-native-image-picker/react-native-image-picker/issues/1532) 

In my case it could happen when user repeat photo select while PickerViewController opened

## Test Plan (required)

1) run app
2) call launchImageLibrary
3) quickly select photo multiple times until PickerViewController is closed
4) repeat launchImageLibrary call and photo select to make sure re-openings work as expected

Feel free to discuss and improve this change

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -30,6 +30,9 @@
 NSString *errCameraUnavailable = @"camera_unavailable";
 NSString *errPermission = @"permission";
 NSString *errOthers = @"others";
+
+BOOL photoSelected = NO;
+
 RNImagePickerTarget target;
 
 RCT_EXPORT_MODULE();
@@ -45,6 +48,7 @@ RCT_EXPORT_METHOD(launchCamera:(NSDictionary *)options callback:(RCTResponseSend
 RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTResponseSenderBlock)callback)
 {
     target = library;
+    photoSelected = NO;
     dispatch_async(dispatch_get_main_queue(), ^{
         [self launchImagePicker:options callback:callback];
     });
@@ -401,6 +405,11 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
 
 - (void)picker:(PHPickerViewController *)picker didFinishPicking:(NSArray<PHPickerResult *> *)results API_AVAILABLE(ios(14))
 {
+    if (photoSelected == YES) {
+        return;
+    }
+    photoSelected = YES;
+
     [picker dismissViewControllerAnimated:YES completion:nil];
 
     if (results.count == 0) {


ios/ImagePickerManager.m
```diff
@@ -30,6 +30,9 @@ @implementation ImagePickerManager
 NSString *errCameraUnavailable = @"camera_unavailable";
 NSString *errPermission = @"permission";
 NSString *errOthers = @"others";
+
+BOOL photoSelected = NO;
```
lukebrandonfarrell, line 34: I'd probably rename this to something a bit more clear like `isPickerOpen` ?

lukebrandonfarrell (changes requested):

ios/ImagePickerManager.m
```diff
@@ -401,6 +405,11 @@ @implementation ImagePickerManager (PHPickerViewControllerDelegate)
 
 - (void)picker:(PHPickerViewController *)picker didFinishPicking:(NSArray<PHPickerResult *> *)results API_AVAILABLE(ios(14))
 {
+    if (photoSelected == YES) {
+        return;
+    }
+    photoSelected = YES;
```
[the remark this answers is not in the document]
Desintegrator: ok i'll gonna check it

ios/ImagePickerManager.m
Desintegrator: no this is not isPickerOpen flag. Cause crash happens if we select photo twice while picker is open. So we just ignore all picks after first one

lukebrandonfarrell: Hmm you mean as in we double tap? I thought this issue was related to opening the picker twice.

ios/ImagePickerManager.m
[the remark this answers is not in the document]
lukebrandonfarrell: Thanks, this will be the case, from my knowledge of the iOS 13 flow

ios/ImagePickerManager.m
Desintegrator: Yep, double tap in my case. Picker is not closing immediate after pick and user (as me) can tap once again

pgrepds:
I think that we can use the same fix to make it work for iOS 13 as well. I've attached a patch file. @lukebrandonfarrell @Desintegrator  and tested this on iOS 15, 14 and 13.

[react-native-image-picker+4.7.3.patch.txt](https://github.com/react-native-image-picker/react-native-image-picker/files/8032556/react-native-image-picker%2B4.7.3.patch.txt)

lukebrandonfarrell:
Merged https://github.com/react-native-image-picker/react-native-image-picker/pull/1948 instead

PR #1928, opened 2022-01-27
fix(android): actions being saved when open launchImageLibrary

arisyo13:

Action popup always show when picking up an image.
When two or more image library apps are installed in users device, then this action popup contains those apps and ask us which one to use. Normally Android takes care of this and asks if you want to save your option. 
Function `createChooser` always enforces this action to being taken.

Here is an example of it 
Screenshot 2022-01-25 at 23 53 29

And here is how it gets when removing `createChooser` function 
Screenshot 2022-01-27 at 15 17 55

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModule.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModule.java
@@ -152,7 +152,7 @@ public class ImagePickerModule extends ReactContextBaseJavaModule implements Act
         }
 
         try {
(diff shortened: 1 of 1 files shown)


lukebrandonfarrell (approved):
LGTM!

slolkunchik:
it helped just for Android 11, not working for Android 12

fadhlinp:
Hi [slolkunchik](https://github.com/slolkunchik), did you solve this problem?

pavelustenko:
still facing this issue

PR #1933, opened 2022-02-04
docs: upgrade example app to react-native 0.66

a7med-mahmoud:
## Motivation

The example project uses react-native 0.63 which doesn't work well with ARM machines (e.g. M1 Macs), so developers who use M1 Macs for development won't be able to run the example app.

## Test Plan

I tested the app on an iOS simulator and an Android emulator.


Johan-dutoit:
Thanks, I'll have a look at this soon and get it merged.

Johan-dutoit (approved):

PR #1955, opened 2022-03-12
feat: add new PhotoPicker API for Android 13 Teramisu

vincent-paing:

## Motivation (required)

Android 13 (codenamed Teramisu) introduces a new [Photo Picker AP](https://developer.android.com/about/versions/13/features/photopicker)I that doesn't require the app to access full media libraries. Going forward, the photo picker will be the recommended way to access the user’s photos and videos. It also provides the API to select a specific amount of image/video to be selected instead of any number.

## Test Plan (required)

- Follow [the doc](https://developer.android.com/about/versions/13/get) to setup Android 13 emulator
- Click Select Image, Select Video or mixed, and check if it pop up the new Photo picker UI
- Play with `0`, `1` and `3` values for `selectionLimit` to see if it correctly handles max, one and certain amount of limit cases.
- Test the same thing with Android < 13 device

(Currently there is known bug where Android 13 emulator does not return correct SDK_INT, so you might need to remove the if/else check for OS version to test it)

*NOTE: It might be too early to merge this PR, this is just a preemptive PR so that we can polish this and merge when Android 13 become stable*

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,11 +1,11 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion rootProject.ext.compileSdkVersion
+    compileSdkVersion "android-Tiramisu"
 
     defaultConfig {
         minSdkVersion 21
-        targetSdkVersion rootProject.ext.targetSdkVersion
+        targetSdkVersion "Tiramisu"
         versionCode 1
         versionName "1.0"
     }
@@ -21,5 +21,5 @@ android {
 dependencies {
     implementation "com.facebook.react:react-native:+"
     implementation "androidx.core:core:1.3.1"
-    implementation "androidx.exifinterface:exifinterface:1.3.1"
+    implementation "androidx.exifinterface:exifinterface:1.3.3"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,11 +1,11 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion "android-Tiramisu"
+    compileSdkVersion 33
 
     defaultConfig {
         minSdkVersion 21
-        targetSdkVersion "Tiramisu"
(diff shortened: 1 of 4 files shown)


android/build.gradle
```diff
@@ -1,11 +1,11 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion rootProject.ext.compileSdkVersion
+    compileSdkVersion "android-Tiramisu"
 
     defaultConfig {
         minSdkVersion 21
-        targetSdkVersion rootProject.ext.targetSdkVersion
+        targetSdkVersion "Tiramisu"
```
Johan-dutoit, line 8: I'm not sure I agree with this change.

lukebrandonfarrell: Yeah this is not a good idea

artyorsh:
> [Android 13 hits Platform stability](https://android-developers.googleblog.com/2022/06/android-13-beta-3-platform-stability.html) and as such, I have updated the sdk versions to the stable version. After you have confirm the feature on pre Android 13 and a Android 13 device/emulator, then feel free to merge it.

@Johan-dutoit given the above, can we have another round of PR review? I also agree it might be early to publish to stable channel, but we can use `@next` or `@beta` tag on npm in order to allow apps to test it (I personally would be happy to participate and give feedback 🙂 )

taher435:
Given the fact that [Android 13 is now available on Pixel devices](https://blog.google/products/android/android-13/), it would be great if this change is available in the next release for an upgrade.

Johan-dutoit (approved):

pierroo:
Quick question though, it seems that this new API no longer grants access to every folders unlike before.
Of course I did add the READ_MEDIA_IMAGES permission in manifest, but still, I can only access favorites / camera / screenshot / downloads, but not every other folders.
It was working fine in earlier versions.

PR #1956, opened 2022-03-12
refactor: Update android build config for latest AGP

vincent-paing:
This also makes the android module easily editable from Android Studio

## Motivation (required)

When I was updating the android module codes, I had to update some configurations to make it work with Android Studio. I noticed that the build script is outdated. I added the buildscript to make it editable in Android studio with zero effort from contributors. I think this will make it a lot easier for contributors to make more PRs.

## Test Plan (required)

- Run CI to make sure the artifact is still generate-able.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,11 +1,21 @@
+buildscript {
+    repositories {
+        google()
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.5.3'
+    }
+}
+
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion rootProject.ext.compileSdkVersion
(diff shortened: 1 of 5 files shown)


android/build.gradle
```diff
@@ -1,11 +1,21 @@
...
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.5.3'
+    }
+}
+
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion rootProject.ext.compileSdkVersion
+    compileSdkVersion 30
```
lukebrandonfarrell, line 14: Why did we remove rootProject.ext.compileSdkVersion instead of just changing the `compileSdkVersion` in `ext` ?

Issue #1977, opened 2022-04-28
[🐛] launchCamera doesn't resolve


Jocodinan:
### Description

Hi!, I have been using react-native-image-picker 4.4.0 for a while without problems on my app, but since I updated to 4.8.1, I have a problem with launchCamera functionality in my iOS device.

When I launch camera for first time, launchCamera response well with the image data, but then, when I launch camera again, it shows the camera, take the photo but I have no response at all and no error throw, just nothing.

```
"react-native": "0.66.4"
"react-native-image-picker": "4.8.1"
```

```
try{
      const result = await launchCamera({
        mediaType: "photo",
        quality: 0.1,
        cameraType: "back",
        includeBase64: true,
        saveToPhotos: true
      });

     console.log(result); //This only works at first time, then it never log a result
  }catch(error){
    console.log("error getting image", error);
  }
```

### How to repeat issue and example

- Just replying the above code

### Solution

- I need that launchCamera functionality always resolve a response when the user take a photo from camera on iOS

### Additional Information

- Image Picker version: 4.8.1
- React Native version: 0.66.4
- Platform: iOS only, Android Works fine
- Development Operating System: MacOS M1 Big Sur 11.3.1
- Dev tools: Xcode 13.2.1

pgrepds:
@lukebrandonfarrell I can confirm this issue, but in my case it doesn't return on the first try. I haven't noticed this since we only use the launchImageLibrary function. I have downgrade to 4.7.3 in which the changes from [#1948 ] are not integrated and the issue does not occur.  I have tested 4.8.0 as well. In short, yes this is very much likely to be caused by [#1948 ].  

A possible hotfix would be to remove the native changes from the iOS side, since the issue we wanted to fix there is very rare. The fix for android was a different issue which is described in [#1944  ] and I think we can leave this change for Android. 

I will create a new PR which rollbacks the changes on iOS, open a new issue which describes the problem the PR fixed for iOS for the launchImageLibrary (but obviously introduced this bug here for launchCamera). This is much cleaner since then we have to separate commits for two different issues. I should have done this initially. I will then look into this new problem. Does this sound good?

@Jocodinan in the meanwhile I would suggest that you downgrade to 4.8.0.

lukebrandonfarrell:
@pgrepds Thanks will be looking out for your PR and we can get it approved

pgrepds:
[...]

PR #1948, opened 2022-02-25
fix(1944): fix runtime exceptions in ImagePickerModule

pgrepds:
## Motivation (required)

We have noticed several crashes due to a RuntimeException for our App on multiple Android devices. The StackTrace given in the Playstore console is as follows.

```
java.lang.RuntimeException:
at android.app.ActivityThread.deliverResults (ActivityThread.java:5471)
at android.app.ActivityThread.handleSendResult (ActivityThread.java:5512)
at android.app.servertransaction.ActivityResultItem.execute (ActivityResultItem.java:51)
at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:149)
at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:103)
at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2386)
at android.os.Handler.dispatchMessage (Handler.java:107)
at android.os.Looper.loop (Looper.java:213)
at android.app.ActivityThread.main (ActivityThread.java:8178)
at java.lang.reflect.Method.invoke (Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:513)
at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1101)
Caused by: java.lang.RuntimeException:
at com.facebook.react.bridge.CallbackImpl.invoke (CallbackImpl.java:26)
at com.imagepicker.ImagePickerModule.onActivityResult (ImagePickerModule.java:184)
at com.facebook.react.bridge.ReactContext.onActivityResult (ReactContext.java:333)
at com.facebook.react.ReactInstanceManager.onActivityResult (ReactInstanceManager.java:785)
at com.facebook.react.ReactDelegate.onActivityResult (ReactDelegate.java:90)
at com.facebook.react.ReactActivityDelegate.onActivityResult (ReactActivityDelegate.java:113)
at com.facebook.react.ReactActivity.onActivityResult (ReactActivity.java:70)
at android.app.Activity.dispatchActivityResult (Activity.java:8413)
at android.app.ActivityThread.deliverResults (ActivityThread.java:5464)
```

We catch the runtime exception and return it via callback.invoke(getErrorMap(errOthers, exception.getMessage())) back to the app where we are able to take further actions and prevent a hard crash.

[...]

PR #1978, opened 2022-05-02
fix(1977): fix reset photo selected on ios

pgrepds:
## Motivation (required)

This PR fixes a bug described in [#1977 ]. This bug was introduced in [this PR](https://github.com/react-native-image-picker/react-native-image-picker/pull/1948).

The variable `photoSelected` was never reset in the case of the function `launchCamera` as it was with the function `launchImageLibrary`.

This issue causes the function `launchCamera` to never resolve on the second call. This PR fixes this.

## Test Plan (required)

Steps to reproduce:

Execute the following twice.

1. Call `launchCamera` and select an image.
2. Console.log the resolved promise either using await or a promise chain.

Prior fix:

- The promise is resolved on first `launchCamera`
- The promise is not resolve on second `launchCamera`

Post fix:
`launchCamera` does resolve as expected on second call (or on every subsequent call of course).

Everything works as expected. The bugs that are resolved in [this PR](https://github.com/react-native-image-picker/react-native-image-picker/pull/1948) are still fixed. The bug in [#1977 ] for `launchCamera` is fixed as well.

I have used the following code for my tests on an iPhone 12 real physical device.

```javascript
async function openCamera() {
    try {
        const result = await launchCamera({
              mediaType: "photo",
              quality: 0.1,
              cameraType: "back",
              includeBase64: true,
              saveToPhotos: true,
         })
        console.log(result)
    } catch (error) {
        console.log(error)
    }
}

<SafeAreaView style={{margin: 20}}>
   <Pressable onPress={openCamera}
        <Text>Open camera</Text>
   </Pressable>
</SafeAreaView>
```

I have performed a regression test for `launchImageLibrary` using a similar workflow.

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -39,6 +39,7 @@ RCT_EXPORT_MODULE();
 RCT_EXPORT_METHOD(launchCamera:(NSDictionary *)options callback:(RCTResponseSenderBlock)callback)
 {
     target = camera;
+    photoSelected = NO;
     dispatch_async(dispatch_get_main_queue(), ^{
         [self launchImagePicker:options callback:callback];
     });


lukebrandonfarrell (approved):
LGTM! Thanks for the fast follow

PR #2001, opened 2022-07-12
feat(ios): add api to dismiss picker modal on ios

tjkang:

## Motivation (required)

When the user runs the app where it left on presentation of image picker through a deep link, the modal of picker should be dismissed. There needs API to dismiss modal manually

## Test Plan (required)
show image picker then when the app goes background, the modal of picker is dismissed

https://user-images.githubusercontent.com/16796578/178428546-c7221b11-cd1d-45fd-8cd0-743a50e3875b.mp4


mtnt:
Why iOS only?

Issue #1894, opened 2021-12-15
IOS: the fileSize of the selected image is bigger then the original one [🐛] 


likashefqet:
### Description

Continues to happen in version 4.

After selecting an image using the library, the asset object contains a fileSize that is different (bigger) from the original image size.
NO maxWidth and NO maxHeight were applied. quality: 1.

This issue has been reported a couple of times in the past [#1419](https://github.com/react-native-image-picker/react-native-image-picker/issues/1419), [#514](https://github.com/react-native-image-picker/react-native-image-picker/issues/514) but these reports have been closed and the issue continues to happen.

### How to repeat issue and example

- Note the image size from the photos app on iOS
- Select that image using this library (NO maxWidth, NO maxHeight, quality: 1) and compare the sizes (sometimes the image is 2 or even 3 times bigger in size)

![IMG_0327](https://user-images.githubusercontent.com/22661589/146235492-e42c7f13-4241-4515-828c-9e51272a3606.jpg)

```javascript
launchImageLibrary({
  includeBase64: false,
  mediaType: 'photo',
  quality: 1,
  selectionLimit: 1,
}, (response) => {
  const { assets } = response;

  if (Array.isArray(assets) && assets.length > 0) {
    const [{ fileSize }] = assets;

    console.log(fileSize / 1024 / 1024);
    console.log(fileSize / 1000 / 1000);
  }
})
```
Expected: 1.9
Result: 
```
6.392001152038574
6.7024989999999995
```

### Solution

What needs to be done to address this issue?

Make sure that the fileSize of the selected image is the same as the original one when no maxWidth and no maxHeight options provided and quality option is set at 1.

### Additional Information
Tested on:
- Image Picker version: v4.1.1 and v4.6.0
- React Native version: v0.63.3
- Platform: iOS (iPhone 11 - real device)
- Development Operating System: MacOS v12.0 and v12.1
- Dev tools:
    - Xcode v13.1 and v13.2
    - iOS v14.4, v15.1 and v15.2
    - node v14.17.4

LefterisL:
Indeed I came up on this myself too. On Android the filesize is reduced just fine but on iOS I once noticed an image from 1.7MB to become 3MB.

christianrank:
iOS stores images internally in High Efficiency Image File Format (.heif / .heic) and converts it to .jpg when it is selected. So it is correct that it gets bigger with quality 1, because jpeg is less efficient.

[...]

PR #2006, opened 2022-07-22
fix: the fileSize of the selected image is bigger than the origin one in ios

ntdiary:

## Motivation (required)
This PR fixes a bug described in [issue 1894](https://github.com/react-native-image-picker/react-native-image-picker/issues/1894).
the binary data returned by `UIImageJPEGRepresentation` and `UIImagePNGRepresentation` is much larger than
the original file.
this PR will use the original data first if user doesn't pass `maxWidth`, `maxHeight` and `quality`

## Test Plan (required)
prepare a image:

1. save the above png image in a device.
2. run example app in the device ( `Team` and `Bundle Identifier` may need to be configured first).
3. click `Select Image` to select the above image
4. compare `fileSize` in the screen with the value in the `Photo` ( or use `ls -l` to check the saved image size)
5. click `Take image` and compare the two values again

In addition,
there is an issue in the `react-native` library when used for upload,
I've provided a temporary solution [here](https://github.com/react-native-image-picker/react-native-image-picker/issues/1894#issuecomment-1191340535)

### DIFF
--- a/ios/ImagePickerManager.m
+++ b/ios/ImagePickerManager.m
@@ -121,20 +121,32 @@ RCT_EXPORT_METHOD(launchImageLibrary:(NSDictionary *)options callback:(RCTRespon
 -(NSMutableDictionary *)mapImageToAsset:(UIImage *)image data:(NSData *)data phAsset:(PHAsset * _Nullable)phAsset {
     NSString *fileType = [ImagePickerUtils getFileType:data];
     
-    if ((target == camera) && [self.options[@"saveToPhotos"] boolValue]) {
-        UIImageWriteToSavedPhotosAlbum(image, nil, nil, nil);
+    if (target == camera) {
+        if ([self.options[@"saveToPhotos"] boolValue]) {
+            UIImageWriteToSavedPhotosAlbum(image, nil, nil, nil);
+        }
+        CFMutableDataRef imageData = CFDataCreateMutable(NULL, 0);
+        CGImageDestinationRef destination = CGImageDestinationCreateWithData(imageData, kUTTypeJPEG, 1, NULL);
+        CGImageDestinationAddImage(destination, image.CGImage, NULL);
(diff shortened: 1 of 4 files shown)


mountiny:
cc @Johan-dutoit hello! I hope we would be able to get this PR merged as we believe it resolves the problem with the increased image size on iOS. Thank you for leaving a review here!

Johan-dutoit (approved):

Johan-dutoit (approved):

mountiny:
Thank @Johan-dutoit for a quick help 🙌

mountiny:
@s77rt can you elaborate on why double the size of original file is an expected output? I think the image picked should have the same size as the image in gallery

s77rt:
@mountiny I think it only seems double because apple stores images in heic format and the conversion from heic to jpeg is not efficient thus the large size.

ios/ImagePickerManager.m
```diff
@@ -121,20 +121,32 @@ - (void) showPickerViewController:(UIViewController *)picker
...
-        data = UIImageJPEGRepresentation(image, [self.options[@"quality"] floatValue]);
-    } else if ([fileType isEqualToString:@"png"]) {
-        data = UIImagePNGRepresentation(image);
+    float quality = [self.options[@"quality"] floatValue];
+    if (![image isEqual:newImage] || (quality >= 0 && quality < 1)) {
+        if ([fileType isEqualToString:@"jpg"]) {
+            data = UIImageJPEGRepresentation(newImage, quality);
+        } else if ([fileType isEqualToString:@"png"]) {
+            data = UIImagePNGRepresentation(newImage);
+        }
```
s77rt, line 149: @Johan-dutoit Sorry for the ping.
Can you take a quick look at this, I think we should revert this part (get ride of the first condition)

Issue #2052, opened 2022-11-18
[🐛] HEIC Files not being converted to JPEG if the quality is set to 1


s77rt:
### Description

HEIC Files not being converted to JPEG if the quality is set to 1

### How to repeat issue and example

- step to reproduce the issue
1. Set the quality to 1
2. Select a heic photo in iOS
3. The returned data is in heic format
4. Set the quality to anything but 1
5. Select a heic photo in iOS
6. The returned data is in jpeg format

In the first case we are using directly the returned data.
In the second case we are using `UIImageJPEGRepresentation` to convert the data. The second case is not relevant here.

To fix the first case we need to get the jpeg representation from iOS but without using `UIImageJPEGRepresentation` because it's so inefficient. Please read the below comments for the fix. 

### Additional Information

- Image Picker version: 4.8.5
- React Native version: 0.70.4
- Platform: iOS
- Development Operating System: MacOS

s77rt:
May be a regression from https://github.com/react-native-image-picker/react-native-image-picker/pull/2036

s77rt:
### Proposal:

on [ImagePickerUtils.m](https://github.com/react-native-image-picker/react-native-image-picker/blob/main/ios/ImagePickerUtils.m)
Change the `preferredAssetRepresentationMode` from [PHPickerConfigurationAssetRepresentationModeCurrent](https://developer.apple.com/documentation/photokit/phpickerconfigurationassetrepresentationmode/phpickerconfigurationassetrepresentationmodecurrent?language=objc) to [PHPickerConfigurationAssetRepresentationModeAutomatic](https://developer.apple.com/documentation/photokit/phpickerconfigurationassetrepresentationmode/phpickerconfigurationassetrepresentationmodeautomatic?language=objc) This setting will automatically select the appropriate format; for heic files we will get jpeg. This is also the behaviour of `UIImagePickerController` (prior iOS 14) so this issue only effect iOS 14+ users due to said problem

This way we will be matching the behaviour of `PHPickerViewController` to `UIImagePickerController`.


Let me know if my proposal make sense and if I should make a PR.

s77rt:
### Alternative fix:
if for any reason changing the `preferredAssetRepresentationMode` is not desirable, we can modify this part of code:
https://github.com/react-native-image-picker/react-native-image-picker/blob/5f2810b2dab2c653b6a1de6cf727cac8aa8553cf/ios/ImagePickerManager.m#L495-L498
[...]

PR #2056, opened 2022-11-29
Change PHPicker preferredAssetRepresentationMode to Automatic

s77rt:
- [X] Target the `main` branch, NOT a "stable" branch.

## Motivation (required)
Selecting HEIC files will return the original format which is in most cases not desired.
Selecting HEIC should get us the JPEG file format.

Changing the `preferredAssetRepresentationMode` from [PHPickerConfigurationAssetRepresentationModeCurrent](https://developer.apple.com/documentation/photokit/phpickerconfigurationassetrepresentationmode/phpickerconfigurationassetrepresentationmodecurrent?language=objc) to [PHPickerConfigurationAssetRepresentationModeAutomatic](https://developer.apple.com/documentation/photokit/phpickerconfigurationassetrepresentationmode/phpickerconfigurationassetrepresentationmodeautomatic?language=objc) will automatically select the appropriate format; for heic files we will get jpeg. This is also the behaviour of `UIImagePickerController`.
 
This way we will be matching the behaviour of `PHPickerViewController` to `UIImagePickerController`.

By the way, before https://github.com/react-native-image-picker/react-native-image-picker/pull/2006 selecting HEIC files was also resulting in JPEG files but this was due to the use of the inefficient `UIImageJPEGRepresentation` conversion.
With this PR, selecting HEIC files will result in JPEG format that is provided by the iOS (no explicit conversion is done from our side)

https://github.com/react-native-image-picker/react-native-image-picker/issues/2052

## Test Plan (required)
The changed code is not that much and I'm not writing new code, I'm not providing a test plan for now as it does not seem to be the case where a test plan is needed, let me know if it's really required.

### DIFF
--- a/ios/ImagePickerUtils.m
+++ b/ios/ImagePickerUtils.m
@@ -58,7 +58,7 @@
         configuration = [[PHPickerConfiguration alloc] init];
     }
     
-    configuration.preferredAssetRepresentationMode = PHPickerConfigurationAssetRepresentationModeCurrent;
+    configuration.preferredAssetRepresentationMode = PHPickerConfigurationAssetRepresentationModeAutomatic;
     configuration.selectionLimit = [options[@"selectionLimit"] integerValue];
 
     if ([options[@"mediaType"] isEqualToString:@"video"]) {


mountiny:
@Johan-dutoit Hello! Would you be able to have a look into this one?

Johan-dutoit (approved):

PR #2057, opened 2022-11-30
[Android] Video uri when pick from galery (file://...)

JerakRus:

- [ ] Target the `main` branch, NOT a "stable" branch.

## Motivation (required)

Recently we found a problem on **Android** devices - **we can't upload the video selected from the gallery** to our server using `formData` (although it used to work). As a result of the study, it turned out that everything works rather strangely. When I take a photo or video with my camera, it comes back with a uri in the format `"file://..."`. When I select a photo from the gallery - uri in the format `"file://..."`. But when I select video from gallery - uri format looks like `"content://..."`. The strangest thing happens when you run `launchImageLibrary({ mediaType: 'mixed', selectionLimit: 0 })`. Then the return result will be an array of image and video files with uri in different non-consistent formats `"file://..."` and `"content://..."`. In addition to the fact that the uri of the `"content://.."` format does not allow uploading the video to the server, it is also not convenient to work with.

**This pull request extends the "file://..." uri creation logic that was already used for photos to video files selected from the gallery. This eliminates this inconsistency and the inability to upload video files to the server.**

## Test Plan (required)

We tested the introduced changes on the Redmi Note 10 Pro device with Android 12. All other cases retained their current behavior. When selecting a video from the gallery, an object with a uri of the format "file://" is returned. The behavior of Ios has not changed. I do not think that there are many problems on other devices, because the same functionality has already been tested for photos. However, testing assistance is welcome. Thank you.

### DIFF
--- a/android/src/main/java/com/imagepicker/Utils.java
+++ b/android/src/main/java/com/imagepicker/Utils.java
@@ -292,7 +292,7 @@ public class Utils {
             case "image/png": return "png";
             case "image/gif": return "gif";
         }
-        return "jpg";
+        return MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType);
     }
 
     static void deleteFile(Uri uri) {
--- a/android/src/main/java/com/imagepicker/Utils.java
+++ b/android/src/main/java/com/imagepicker/Utils.java
@@ -447,6 +447,7 @@ public class Utils {
         for(int i = 0; i < fileUris.size(); ++i) {
             Uri uri = fileUris.get(i);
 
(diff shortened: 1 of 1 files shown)


Johan-dutoit:
👋  Thanks for the contribution.  This seems like a major version update (as it appears to be breaking).
Thoughts?

Otherwise, I think it makes sense.

infoSamyak:
Hello,
This PR does fixes an issue which a lot developers are facing. Its your call.
I just request you to merge this PR ASAP.
Thank You

Johan-dutoit (approved):

PR #2062, opened 2022-12-15
fix(types): Define CameraOptions and ImageLibraryOptions symmetrically

chrisbobbe:
By factoring out an `interface OptionsCommon` that both interfaces extend, instead of having one extend the other minus a property. This seems cleaner, since the properties are now neatly organized into three groups:

- in common
- unique to ImageLibraryOptions
- unique to CameraOptions

This doesn't actually change either type, just how we structure their definitions, so I don't expect it to add or remove any type errors that TypeScript reports.

It also makes the TypeScript file a bit friendlier to conversion to Flow via TsFlower: https://github.com/gnprice/tsflower

-----

## Motivation (required)

(See above)

## Test Plan (required)

Could we add a `yarn test` script that runs the TypeScript type checker? I'm not very familiar with how to operate TS (my project uses Flow for its commitment to soundness), so my test plan at this point is just to make sure CI passes.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,7 +1,6 @@
 export type Callback = (response: ImagePickerResponse) => any;
 
-export interface ImageLibraryOptions {
-  selectionLimit?: number;
+export interface OptionsCommon {
   mediaType: MediaType;
   maxWidth?: number;
   maxHeight?: number;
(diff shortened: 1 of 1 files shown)


Johan-dutoit (approved):
Totally for this change, thanks!

src/types.ts
```diff
@@ -1,7 +1,6 @@
 export type Callback = (response: ImagePickerResponse) => any;
 
-export interface ImageLibraryOptions {
-  selectionLimit?: number;
+interface OptionsCommon {
```
Johan-dutoit, line 3: I'd probably export this too.

chrisbobbe:
Thanks for the review! Revision pushed.

Johan-dutoit (approved):

PR #2094, opened 2023-03-13
feat(ios): Add feature to convert videos to MP4

Johan-dutoit:
This allows videos to be converted to MP4 on iOS.  For now the quality is lowered, to make it faster.

Non-MP4 video (no conversion)
image

MP4 video (same video as above, with conversion)
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,22 +90,21 @@ The `callback` will be called with a response object, refer to [The Response Obj
 
 ## Options
 
-| Option         | iOS | Android | Web | Description                                                                                                                         |
-| -------------- | --- | ------- | --- |------------------------------------------------------------------------------------------------------------------------------------ |
-| mediaType      | OK  | OK      | OK  | 'photo' or 'video' or 'mixed'(launchCamera on Android does not support 'mixed'). Web only supports 'photo' for now.                             |
-| maxWidth       | OK  | OK      | NO | To resize the image                                                                                                                       |
-| maxHeight      | OK  | OK      | NO | To resize the image                                                                                                                       |
-| videoQuality   | OK  | OK      | NO | 'low', 'medium', or 'high' on iOS, 'low' or 'high' on Android                                                                             |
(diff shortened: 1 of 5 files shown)


ratz6:
@Johan-dutoit  @marcshilling @dwosk @MakarovAleksey 

can we add this condition to run if the file extension is not .mp4 ? 

It's running even if the fileType is .mp4 when formatAsMp4 is passed as true .. 
so the copying to temp directory is happening twice for both .mov as well as .mp4 video.

Johan-dutoit:
Good point @ratz6 , can you submit a PR please to fix it?

ratz6:
@Johan-dutoit is this ok ? 
https://github.com/react-native-image-picker/react-native-image-picker/pull/2133
Haven't had a chance to get my hands dirty with objective C. Please don't mind :)

ratz6:
> Good point @ratz6 , can you submit a PR please to fix it?

@Johan-dutoit @marcshilling @dwosk @MakarovAleksey is this ok ?
https://github.com/react-native-image-picker/react-native-image-picker/pull/2133
Haven't had a chance to get my hands dirty with objective C. Please don't mind :)

ratz6:
@Johan-dutoit there are few cases not handled while formattingToMp4 , I'm fixing it in swift but pls don't ask me to raise PR in obj c .. Is there a scope to migrate whole package to swift ? > 

1) Unable to read files with spaces. in the device. Convert spaces from fileName into "" when copying the file into the app's temp memory .. 
Did this in swift > 

` var temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(String(url.lastPathComponent!.filter { !" \n\t\r".contains($0) }) )`

2) If I don't pass formatAsMp4 as true , the the file is being copied to temp folder by taking the path which ends with fileName (coming from device) as three is no UUID appended here.. this file is already deleted as the newly picked file will also go to the same place
image

PR #2118, opened 2023-04-20
Add fallback to getting extension from filename

j-piasecki:

## Motivation (required)

In some cases the [`getExtensionFromMimeType`](https://android.googlesource.com/platform/frameworks/base/+/0ea8c8a7c060873ef94e8aa8a637bc90852777d1/core/java/android/webkit/MimeTypeMap.java#151) may return `null`, in which case the generated file would lack the extension. This PR adds fallback to rely on the extension when it happens.

## Test Plan (required)

I wasn't able to make a reproducible example to showcase the problem, however, it emerged during the migration of the Expensify app to the new architecture. [We've been using the patch](https://github.com/Expensify/App/blob/a9c5bcbf52ee7c368c5390d58ea489d6b4f53d6e/patches/react-native-image-picker%2B5.3.1.patch#L25-L39) for a while and didn't see any problems with it.

### DIFF
--- a/android/src/main/java/com/imagepicker/Utils.java
+++ b/android/src/main/java/com/imagepicker/Utils.java
@@ -8,6 +8,7 @@ import android.content.ContentValues;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.hardware.camera2.CameraCharacteristics;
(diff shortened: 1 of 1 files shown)


Johan-dutoit:
@j-piasecki keen to update the PR to also grab the filename from the content resolver?

j-piasecki:
@Johan-dutoit could you clarify what you mean? It already gets the filename via `ContentResolver.query(...)`, no?

Johan-dutoit:
Hah.  Apologies for not being clear.  I was suggesting to use the filename from the `ContentResolver.query(...)` to the output (e.g. replacing the current `String fileName = uri.getLastPathSegment();` with the value from ContentResolver.

Johan-dutoit:
Actually, I'll make the change once this PR is merged.  Thanks!

Johan-dutoit (approved):

Issue #2138, opened 2023-06-07
[🐛] Android: Application not responding after selecting many videos


renchap:
### Description

When selecting many videos in the new Android 13 picker, an Android "Application Not Responding" message appears.

If "wait" is pressed, then the application works fine, so the main thread is not the one that is stuck.

Detail about the videos is shown, and thumbnails are here as expected.

### How to repeat issue and example

- launch the example app on an Android 13 phone
- tap "select image or video"
- select 15 videos from your library
- wait for the videos to be process, then see that an ANR dialog is shown

### Solution

There should not be any ANR

### Additional Information

- Image Picker version: 5.4.0 (also 5.3.1)
- React Native version: 0.71.2 (example app), 0.71.8 (my app)
- Platform: Android
- Development Operating System: macOS

renchap:
I think this is fixed by #2139, I am no longer reproducing the issue with this patch.

renchap:
Update: I have been testing this further, and while the issue occurs less often, I still get ANRs when selecting a lot of videos using the new Android 13 photo picker.

I am starting to suspect this is an Android issue.

renchap:
Still investigating this…

It only happens when `getAppSpecificStorageUri()` is called, meaning the files are copied to the app's cache directory.

This led to #2141 which fixes some issues, but unfortunately I can still reproduce the crashes.

I tried to not copy the files to the app's cache directory, but instead require Android to make the `content://` URLs persistant, and this seems to solve the issue. I will tidy up my work on this and submit a PR.

renchap:
Another update:
- `MediaMetadataRetriever` also needs to be closed after use, I updated #2141 to do it. This change made a big difference and after some testing I did not get ANRs once the files were processed
- The processing is done in the main thread, which can cause it to be unresponsive and trigger ANRs while the files are processed after being picked. This also makes the app UI non-responsive. Opened #2142 to run the processing in another thread

With those 2 changes I am no longer experiencing ANRs or crashed, even when selecting 50 videos on the phone.
`getAppSpecificStorageUri()` still makes things quite slow as every file needs to be copied in the app's storage, so I may still send yet another PR to allow avoiding this step through an option.

renchap:
Closing the issue as the PRs mentioned above have been merged and fixes the issues for me.

PR #2141, opened 2023-06-08
Fix android leaks

renchap:
## Motivation (required)

While working on solving #2138, @huextrat noticed that the various streams and descriptors were never closed.

This is leaking a lot of resources.

I converted those to try-with-resources syntax, so they are automatically closed after the block ends.

## Test Plan (required)

I tested various files and videos with the example app, and did not notice any issue.

I also enabled `StrictMode` on the example app (see https://wh0.github.io/2020/08/12/closeguard.html) and it no longer detects non-closed resources.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImageMetadata.java
+++ b/android/src/main/java/com/imagepicker/ImageMetadata.java
@@ -8,8 +8,7 @@ import java.io.InputStream;
 
 public class ImageMetadata extends Metadata {
   public ImageMetadata(Uri uri, Context context) {
-    try {
-      InputStream inputStream = context.getContentResolver().openInputStream(uri);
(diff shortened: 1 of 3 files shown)


renchap:
This has now been tested on Android 9 as well.

`MediaMetadataRetriever` does not implements `AutoCloseable` on API < 29, so I added a custom wrapper.

I also removed my commit replacing the byte-copy loops with `InputStream.transferTo()` as this somehow does not work on Android 9 (code execution stops there and the JS Promise never returns).

Johan-dutoit (dismissed):

Johan-dutoit:
@renchap please resolve conflicts.  Will review/merge then.

Johan-dutoit (approved):

PR #2145, opened 2023-06-16
feat(android): permit to specify a list of allowed mime-types

renchap:
## Motivation (required)

The Android camera app has an option to store both JPEG and RAW (DNG format) files for each picture. When this option is enabled, both versions of the image are shown to the user (but without any indication of the file format, so the users sees every image twice).

This new options allows you to restrict the allowed MIME types, for example `restrictMimeTypes: ['image/jpeg', 'image/heic', 'image/png']`.

On Android 13 (new Photo Picker), only the matching files are displayed, no more duplicates.

On older Android versions, every file is still displayed in the file explorer, but non-matching files are grayed and can not be selected.

This option can not be implemented on iOS, as the API only allows you to filter on predefined types (images, videos, slow-mos…).

## Test Plan (required)

I used `restrictMimeTypes: ['image/jpeg', 'image/heic', 'image/png']` with the example app and tested the behaviour with and without the option on an Android 10 and Android 13 devices.

### DIFF
--- a/android/src/main/java/com/imagepicker/ImagePickerModuleImpl.java
+++ b/android/src/main/java/com/imagepicker/ImagePickerModuleImpl.java
@@ -157,6 +157,10 @@ public class ImagePickerModuleImpl implements ActivityEventListener {
             libraryIntent = pickMultipleVisualMedia.createIntent(this.reactContext.getApplicationContext(), mediaRequest);
         }
 
+        if(this.options.restrictMimeTypes.length > 0) {
+            libraryIntent.putExtra(Intent.EXTRA_MIME_TYPES, this.options.restrictMimeTypes);
+        }
+
         try {
             currentActivity.startActivityForResult(libraryIntent, requestCode);
         } catch (ActivityNotFoundException e) {
--- a/example/src/components/DemoResponse.tsx
+++ b/example/src/components/DemoResponse.tsx
@@ -7,7 +7,7 @@ export function DemoResponse({children}: React.PropsWithChildren<{}>) {
   }
 
   return (
-    <ScrollView style={styles.container}>
+    <ScrollView style={styles.container} nestedScrollEnabled>
       <Text style={styles.text}>{JSON.stringify(children, null, 2)}</Text>
     </ScrollView>
   );
--- a/src/platforms/native.ts
+++ b/src/platforms/native.ts
@@ -9,6 +9,7 @@ import {
 
 const DEFAULT_OPTIONS: ImageLibraryOptions & CameraOptions = {
   mediaType: 'photo',
+  restrictMimeTypes: [],
   videoQuality: 'high',
   quality: 1,
   maxWidth: 0,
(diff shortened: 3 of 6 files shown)


android/src/main/java/com/imagepicker/ImagePickerModuleImpl.java
```diff
@@ -125,35 +125,42 @@ public void launchImageLibrary(final ReadableMap options, final Callback callbac
         boolean isPhoto = this.options.mediaType.equals(mediaTypePhoto);
         boolean isVideo = this.options.mediaType.equals(mediaTypeVideo);
 
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+        boolean usePhotoPicker = Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU;
```
renchap, line 128: I moved this check into a variable as it was used multiple times

android/src/main/java/com/imagepicker/ImagePickerModuleImpl.java
```diff
@@ -125,35 +125,42 @@ public void launchImageLibrary(final ReadableMap options, final Callback callbac
...
+
+        if (usePhotoPicker) {
```
renchap, line 130: Order of the statements has been inversed to avoid a negation in the `if` statement

example/src/components/DemoResponse.tsx
```diff
@@ -7,7 +7,7 @@ export function DemoResponse({children}: React.PropsWithChildren<{}>) {
   }
 
   return (
-    <ScrollView style={styles.container}>
+    <ScrollView style={styles.container} nestedScrollEnabled>
```
renchap, line 10: Without this option, the view is not scrollable on Android and you only see the first lines

src/platforms/native.ts
```diff
@@ -26,9 +27,9 @@ const DEFAULT_OPTIONS: ImageLibraryOptions & CameraOptions = {
 // @ts-ignore We want to check whether __turboModuleProxy exitst, it may not
 const isTurboModuleEnabled = global.__turboModuleProxy != null;
 
-const nativeImagePicler = isTurboModuleEnabled ?
-  require("./NativeImagePicker").default :
-  NativeModules.ImagePicker;
+const nativeImagePicler = isTurboModuleEnabled
```
renchap, line 30: This has been formatted by Prettier, this file was not up-to-date

renchap:
@Johan-dutoit Could you have a look at this PR?

It has been running in our production app for 2 months without issues.

Julesssss:
> @Johan-dutoit Could you have a look at this PR?
> It has been running in our production app for 2 months without issues.

Bumping this comment

renchap:
Rebased the PR against `main`, and tested that it works fine on a Pixel 5 phone.

Johan-dutoit (approved):

PR #2148, opened 2023-06-19
feat(android): 🌟 add support for React Native 0.73

Johan-dutoit:
As per https://github.com/react-native-community/discussions-and-proposals/issues/671, this is to support a future release of React Native

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,6 +8,7 @@ if (isNewArchitectureEnabled()) {
 }
 
 android {
+    namespace = "com.imagepicker"
     compileSdkVersion 33
 
     // Used to override the NDK path/version on internal CI or by allowing
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest
   xmlns:android="http://schemas.android.com/apk/res/android"
-  package="com.imagepicker"
   >
     <application>
       <provider


markrickert:
There's a way you can maintain support for older versions of react native. Check out how react-native-gesture-handler did it: https://github.com/software-mansion/react-native-gesture-handler/pull/2448/files

tadjik1:
@Johan-dutoit do you need help with that? I'm stuck on old version of RN and because of that can't build module

jeffinhk:
> @markrickert thanks for that! Will update shortly!

it's been few months, so any updates on this?
for those who are looking for building issues with this, you can first try the work around :

add back the 
  package="com.imagepicker"
in AndroidManifest.xml of this dependency in your node_modules.

Here is what is should look like :

<?xml version="1.0" encoding="utf-8"?>
  <manifest
    xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.imagepicker"
   >
...

ivanignatiev:
Removal of package in manifest is optional.

You also need to add `android.buildFeatures.buildConfig true` otherwise it will fail:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':react-native-image-picker'.
> com.android.builder.errors.EvalIssueException: defaultConfig contains custom BuildConfig fields, but the feature is disabled.
  To enable the feature, add the following to your module-level build.gradle:
  `android.buildFeatures.buildConfig true`

BUILD FAILED in 4s
error Failed to install the app.
```

PR #2157, opened 2023-06-21
feat(ios): 🌟 ability to trigger limited permissions picker

Johan-dutoit:
This provides iOS users with the ability to update which photos the app has access to.


Johan-dutoit:
Closing. [react-native-permissions](https://github.com/zoontek/react-native-permissions) already covers support for this.

harbolaez:
@Johan-dutoit how? They only support updating the selected images but not showing those limited selected one?

Issue #2092, opened 2023-03-02
[🐛] Exif/Meta data of selected/taken images not being extracted on iOS


jsimmons12:
### Description

On iOS, no meta/exif data is being returned on the images selected or taken using the react-native-image-picker library. I have the 'includeExtra' option enabled and it is working correctly on Android, however on iOS it appears as though all of the meta data is being lost. 

On images taken with the device, no meta data is captured at all. Using react-native-fs I have saved the images taken using react-native-image-picker directly to the device and copied them to my computer to look at in more detail, which reveals the files themselves have no meta data attached to them.

On images already on the device that have been taken using the device's primary camera, no meta data is being harvested despite it being visible when examining the files themselves. 

Appears to be an issue with the process of copying the selected image to the temp directory, at which point the meta data is being lost.

### How to repeat issue and example

My code for selecting images/videos:

```
  selectMedia: async function (type: MediaType) {
    const result = await launchImageLibrary({
      mediaType: type,
      includeExtra: true,
      selectionLimit: 0,
      videoQuality: 'medium'
    });

    return result.assets;
  }

```

My code for taking images/videos:

```
  openCamera: async function (type: MediaType, platform: PlatformOSType) {

    let options: CameraOptions = {
      mediaType: type,
      includeExtra: true,
      cameraType: 'back',
      videoQuality: 'medium',
    }

    if (platform === 'ios') {
      options.maxWidth = 4032;
      options.maxHeight = 3024;
    }

    const result = await launchCamera(options)

    if (result.assets) {
      return result.assets;
    }
  }
```

### Solution

Capture the correct meta data and ensure it's not being lost in the file creation process. 

### Additional Information

- Image Picker version: 5.0.2
- React Native version: 0.65.1
- Platform: iOS
- Development Operating System: MacOS
- Dev tools: Xcode  14.2, iOS 16.3.1

mehdinourollah:
Any update ?!

ArturoTorresMartinez:
Same here

cristian0791:
Any update?

SayisMe:
Same here too

NomanGul:
I applied this patch to get full exif data on iOS (including GPS info)

```patch
diff --git a/node_modules/react-native-image-picker/ios/ImagePickerManager.mm b/node_modules/react-native-image-picker/ios/ImagePickerManager.mm
index 93e99be..da8d721 100644
[...]

PR #2165, opened 2023-06-27
fix: add support for iOS missing exif data

mechazod:
Issue with photo export not adding existing `exif data` which vital aspect of an image.

Many issues [raise](https://github.com/react-native-image-picker/react-native-image-picker/issues?q=is%3Aissue+is%3Aopen+exif)


mo-nathan:
I note that this PR reference version 5.6.0, but the latest version is 7.1.2.  I think I'll try the patch discussed in #2092.

PR #2209, opened 2023-09-07
Retrieve the original filename on android

Johan-dutoit:
BREAKING CHANGE:  changed the way the name is retrieved


android/src/main/java/com/imagepicker/Utils.java
```diff
@@ -396,22 +395,65 @@ static String getMimeType(Uri uri, Context context) {
...
+        }
+
+        return "Unknown";
+    }
+
+    static String getOriginalFilePath(Uri uri, Context context) {
+        String originPath;
+        if (uri.getScheme().contains("content")) {
+            originPath = getFilePathFromContent(uri, context);
+            uri = getAppSpecificStorageUri(uri, context);
```
mattlewer, line 423: This causes a second cached file to be created which isn't returned. Issue created here: https://github.com/react-native-image-picker/react-native-image-picker/issues/2271

PR #2221, opened 2023-09-29
Update iOS image picker

### LINKED ISSUE #2220: [🐛]
### Description

A device in iOS 16 that captures a photo and syncs to iCloud Photo Library that an iOS 17 app selects with this library fails the check for PHPhotoPicker when loading it he UIImage class. 

I plan to fix this in our app, patch the library and submit a PR. 

Explain the bug in as much detail as possible

- see description 

### Solution

What needs to be done to address this issue?

I will update the API to use the item for identifier instead of data for class. 

### Additional Information

- Image Picker version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
- React Native version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
- Platform: [FILL THIS OUT: iOS, Android, or both?]
- Development Operating System: [FILL THIS OUT: Are you developing on MacOS, Linux, or Windows?]
- Dev tools: [FILL THIS OUT: Xcode or Android Studio version, iOS or Android SDK version, if applicable] navigation crash for Android

jefflewis:
* Preserve metadata for images when targeting JPEG output
* Fixes #2220 - continue picker if only available format is HEIF or HEIC

## Motivation (required)

This change has two motivations:

* preserve image metadata instead of tossing it away
* workaround what appears to be an API bug in iOS17 with `NSItemProvider#canLoadObjectOfClass` returning `false` for an item provider with only a `HEIF` or `HEIC` image.

## Test Plan (required)

For the iCloud but, this requires testing with two live devices to verify things work as expected.
1.  Take a photo on a device logged into an iCloud account.
2. Use this library on  a device on iOS 17 logged into the same iCloud account.
3. Select the image synced over with iCloud (without making any edits)
4. Without this change, see that the results array is empty. With this change, see that you have your image.

For the metadata:
1. Select any image. 
2. View the metadata of the image after selection before and after this change. Before, metadata like EXIF and GPS is missing. With this change, metadata is included. (both with camera or picker -- GPS would require a separate effort, but image EXIF data will be included.)

### DIFF
--- a/ios/ImagePickerManager.mm
+++ b/ios/ImagePickerManager.mm
@@ -1,3 +1,4 @@
+#import "UniformTypeIdentifiers/UniformTypeIdentifiers.h"
 #import "ImagePickerManager.h"
 #import "ImagePickerUtils.h"
 #import <React/RCTConvert.h>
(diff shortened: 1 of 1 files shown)


mglobel:
Hey @jefflewis thank you for this! I tried this fix out and it did attach photos that had been synced from another iCloud device. However the orientation was off (always landscape instead of portrait). Are you experiencing that issue?

jefflewis:
Oh interesting. We've always had rotation issues, and have a second process that rotates the image to the specified exit tag value. With that, we've had no troubles before or after this change. 

Can you inspect the orientation value of the exit and see if it is correct on the image you see with incorrect rotation?

efstathiosntonas:
@jefflewis thanks for this fix. 

I'm getting lots of errors though on rn 0.72.5 and picker@7.0.1

Screenshot 2023-10-11 at 10 00 52

Screenshot 2023-10-11 at 10 00 46

Screenshot 2023-10-11 at 10 00 21

jefflewis:
> @jefflewis thanks for this fix.
> 
> I'm getting lots of errors though on rn 0.72.5 and picker@7.0.1
> 
> Screenshot 2023-10-11 at 10 00 52
> 
> Screenshot 2023-10-11 at 10 00 46
> 
> Screenshot 2023-10-11 at 10 00 21
> 
> this method is not implemented anywhere in the code: `- (id)mapImageToAsset:(UIImage *)image data:(NSData *)data phAsset:(PHAsset *)phAsset metadata:(NSDictionary *)metadata;`

You are right! I misapplied a local patch when I pushed this my branch here. I have updated it with the correct method.

efstathiosntonas:
@jefflewis thanks for the update, it still throws these errors: 

Screenshot 2023-10-11 at 20 30 22


Screenshot 2023-10-11 at 20 30 16

charlyBerthet:
any news on this PR ?

Issue #2119, opened 2023-04-21
[🐛] Error processing request body: Could not find image file


dmsierra11:
### Description

Error when trying to upload image from gallery

![Simulator Screen Shot - iPhone 12 - 2023-04-20 at 13 10 32](https://user-images.githubusercontent.com/2533919/233573856-6e5f8e33-36b6-4e06-a1bf-8b5ffb2a14d0.png)

### How to repeat issue and example

```
const imagePickerDefaultOptions = {
  selectionLimit: 0,
  mediaType: 'photo' as MediaType,
  includeBase64: false,
  includeExtra: true,
};

launchImageLibrary(imagePickerDefaultOptions, (imgPickerResponse) => {
    if (imgPickerResponse.assets?.[0]) {
      onPhotoTaken(imgPickerResponse.assets?.[0]);
    }
  });

```
### Solution

What needs to be done to address this issue?

### Additional Information

- Image Picker version: 4.10.3
- React Native version: 0.66.1
- Platform: iOS
- Development Operating System: MacOS
- Dev tools: Xcode 14.0.1, iOS 16.0

yorickshan:
same issue

VanDinh18:
same issue

cokron:
Same here - upgraded from react-native-image-picker 4.10.3 to  5.3.1. 
It used to work in 4.10.3 both in Simulator and on real device (iPhone).
Now it stopped working. Going back to 4.10.3 for now.

jeremiahlachica:
I also have the same issue. Currently using 5.3.1

maydersonmellops:
Same error here, until this correction was made, I used lib `react-native-image-crop-picker` and implemented it as follows:

```tsx
import ImageCropPicke from 'react-native-image-crop-picker';

ImageCropPicker.openPicker({
  mediaType,
   cropping: false,
   compressImageQuality: 0.8,
})
   .then(data => console.log(data))
   .catch(error => console.log(error));
```

Had a similar result to Image Picker.

maydersonmellops:
@dmsierra11 I believe the problem is here.

Successfully selected image:

```json
{
  "fileName":"5693202E-E952-4C0D-A69B-2FF9B0F6C688.jpg",
  "fileSize":993045,
  "height":2500,
  "type":"image/jpg",
[...]

PR #2236, opened 2023-11-09
fix(#2119): make live photo identifier more generic

samcodefish:
## Motivation

[Issue 2119](https://github.com/react-native-image-picker/react-native-image-picker/issues/2119)

We are running into the problem outlined in issue 2119 where there is a "Could not find image file..." error on iOS when selecting images that are on iCloud but not yet fully synced to the user's device.

I found by updating the live photo identifier to be more generic (rather than it explicitly being "public.jpeg") this resolved the issue.

_Please note: I am a React Native / Javascript developer, and not by any means a native iOS / Objective C developer. To reach the solution in this PR I referenced the code in [this PR](https://github.com/react-native-image-picker/react-native-image-picker/pull/2221), and added the least amount I could to resolve the issue. Someone with more native iOS experience may be able to suggest a more robust solution to the issue._

## Test Plan

(Requires testing with two devices logged into the same iCloud account)

- Take a photo on device A
- Use the example app on device B logged into the same iCloud account
- Select the image synced over with iCloud (without making any edits)
- Without this change you'll get the "Could not find image file..." warning. With this change, you'll retrieve the image as expected.

### DIFF
--- a/ios/ImagePickerManager.mm
+++ b/ios/ImagePickerManager.mm
@@ -1,3 +1,4 @@
+#import "UniformTypeIdentifiers/UniformTypeIdentifiers.h"
 #import "ImagePickerManager.h"
 #import "ImagePickerUtils.h"
 #import <React/RCTConvert.h>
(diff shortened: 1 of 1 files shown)


efstathiosntonas:
Hi folks, any updates on this one?

Johan-dutoit (approved):

PR #2245, opened 2023-11-23
Example with notifee foreground notification to keep the app alive in background

dprevost-LMI:

## Motivation (required)
I enhanced the example project to give a complete, working example of the Notifee foreground service, keeping the app alive in the background when taking a picture or video with the camera.

Solve the problem where the app is killed in the background while using the camera, and when exiting the camera, your app gets restarted and lands on the home page.

## Test Plan (required)

Using the simulator on different Android versions, I validated that we were seeing the foreground notification.
This is also where I found that Android 8 and 9 required camera permission, even if it was not in the manifest. , So I provide another example of requesting camera permission.

Screenshot 2023-11-22 at 10 42 48 PM
Screenshot 2023-11-22 at 10 42 56 PM
Screenshot 2023-11-22 at 10 43 21 PM
Screenshot 2023-11-22 at 10 43 44 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ example/ios-bundle.js
 index.android.bundle
 index.ios.bundle
 lib/
+.history/*
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -14,5 +14,4 @@
                 android:resource="@xml/imagepicker_provider_paths" />
         </provider>
     </application>
-    <uses-permission android:name="android.permission.CAMERA" />
 </manifest>
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -14,4 +14,5 @@
                 android:resource="@xml/imagepicker_provider_paths" />
         </provider>
     </application>
+    <uses-permission android:name="android.permission.CAMERA" />
 </manifest>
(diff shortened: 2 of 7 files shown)


dprevost-LMI:
@gabrieldonadel or @cipolleschi you seem to commit often. Is the above has value for this project else I will just close it

Wesleyomnion:
@dprevost-LMI , 
Thank you very much my friend, I didn't know what to do, your solution worked very well for me, thank you very much. Success.

dprevost-LMI:
If only we could merge it 😉

ksanbormukhim:
still using @dprevost-LMI solution and still working. Is this the going to be the "temporary" fix for this problem or have we found a better solution.

dprevost-LMI:
Since this involves another activity/intent, I do not think there is a better solution. To avoid having to use this, we need to embed the camera in the app instead, using other libraries, but that would be much more work.

Sahilr2161:
@dprevost-LMI How are you handling Foreground service permissions on the Play Store?

dprevost-LMI:
> @dprevost-LMI How are you handling Foreground service permissions on the Play Store?

Since we are a VOIP and meeting application, it was easy to justify. We used foreground service with the camera.

If you say you need to perform processing on the camera in the foreground service while taking a picture, that would be justifiable.

PR #2248, opened 2023-11-29
Android 13 permission fixes

AdityaPrakash04:


leon8941:
Is this still required? If not we can close this PR instead letting it sit here.

PR #2264, opened 2024-01-13
feat(android): convert HEIC to JPEG

khitrin:
BREAKING CHANGE: automatically convert HEIC files to JPEG on Android if assetRepresentationMode is not set to current

## Motivation (required)

Some Android phones (Samsung for example) take photos in HEIC format like iPhones.
The library converts HEIC to JPEG on iOS using preferredAssetRepresentationMode, but the Android picker seems to lack this ability.
Sharing unconverted HEICs is very inconvenient because many clients cannot display them.

The change adds an ability to the library to mimic the iOS behaviour and convert selected HEIC files to JPEG on Android.
Support for preferredAssetRepresentationMode option has been added to Android to provide an opt-out for this conversion working exactly the same as on iOS. If a user wants to select the original file they can set it to "current" as they already do on iOS.

Also conversionQuality option has been added to set the desired quality of converted images because existing quality option must be set to 1 to avoid JPEG-to-JPEG conversion. Default value 92 is chosen as same to the ImageMagick defaults.

## Test Plan (required)

1. Add a HEIC file to the gallery (or take a photo on Samsung).
2. Select image using unchanged Example App. 
before
3. Type of the image is "image/heic", extension of uri is ".heic".
4. Apply patch, recompile and update the app, select the same image. 
after
5. Type of the image is "image/jpeg", extension of uri is ".jpg", fileSize is slightly increased.


SveBo (approved):
Thanks, this is a short and good solution, works perfectly.

SveBo (approved):

sjchristi:
Having issues with newer samsung phones saving in HEIC; would love to see this integrated.

android/src/main/java/com/imagepicker/Options.java
```diff
@@ -10,6 +10,8 @@ public class Options {
     Boolean includeExtra;
     int videoQuality = 1;
     int quality;
+    int conversionQuality = 92;
```
Johan-dutoit, line 13: What's the reason for 92 as the defautl?

khitrin: Same as ImageMagick defaults.

https://imagemagick.org/script/command-line-options.php#quality

sjchristi:
been using khitrin's branch in production and it's been fine.

now we're being forced to update our android to use the photo picker.  could someone please merge this?

Johan-dutoit (approved):

Issue #2318, opened 2024-08-12
[🐛] undefined symbol


EzequielMonteleone:
### Description
I am receiving this error in iOS, in android work fine
 "react-native": "0.74.2"
 "react-native-image-picker": "^7.1.2"
<img width="1792" alt="Screenshot 2024-08-12 at 1 36 17 PM" src="https://github.com/user-attachments/assets/d3c05617-1132-40a7-9b39-66cc447b8cd7">

Explain the bug in as much detail as possible

### How to repeat issue and example

- step to reproduce the issue
- ...

### Solution

What needs to be done to address this issue?

### Additional Information

- Image Picker version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
- React Native version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
- Platform: [FILL THIS OUT: iOS, Android, or both?]
- Development Operating System: [FILL THIS OUT: Are you developing on MacOS, Linux, or Windows?]
- Dev tools: [FILL THIS OUT: Xcode or Android Studio version, iOS or Android SDK version, if applicable] navigation crash for Android

jpancotti:
Same problem with react native image picker 7.1.2 on both react native 73.9 and 74.0

EzequielMonteleone:
4.0.6 work!

TheNoShade:
I have same issue any solution?

jguerinet:
@TheNoShade What I have done temporarily is change it to be a static framework so that it includes everything. You can do so by adding this to your Podfile: 
```
pre_install do |installer|
    installer.pod_targets.each do |target|
      # Make some targets static as they throw errors if not
      if target.name.eql?('react-native-image-picker')
        def target.build_type;
          Pod::BuildType.static_library
        end
      end
    end
  end
```

TheNoShade:
<img width="384" alt="Screenshot 2024-09-10 at 12 24 47 PM" src="https://github.com/user-attachments/assets/7fba9e77-d904-4600-9d8c-bc24dcbb459b">

Now facing this error

jguerinet:
This is for another library, so you will need to see if there's a fix for that or maybe also include this library in the script above to be build as a static library.

Johan-dutoit:
This should be fixed in newer versions, please check

PR #2320, opened 2024-08-30
fix(ios): missing required frameworks

decim92:

## Motivation (required)
After installing version 7.1.2, when I run the iOS project it won't compile since it complains of undefined classes
It is related to this issue https://github.com/react-native-image-picker/react-native-image-picker/issues/2318

It compiles the iOS project in latest version

## Test Plan (required)
I don't know how you make it work in your environment but if these frameworks are missing it won't work


Johan-dutoit:
Interestingly, the example app (freshly created) builds fine when using the latest image picker.

Issue #2364, opened 2025-04-07
[🐛] Android - Exif GPS data is empty & exif data gets lost


DylanJCraggs:
### Description

When selecting an image from the image gallery on android, the EXIF data does not come down from the image due to the PickMultipleVisualMedia stripping it (see https://issuetracker.google.com/issues/243294058).

Also another issue was found after resolving that where resizing the image would then wipe all the EXIF data (including the GPS exif data).

### How to repeat issue and example
On Android, launch the image gallery with the following options:
{
        quality: 0.1,
        presentationStyle: 'fullScreen',
    };

Select an image that has GPS exif data, and then in the device explorer, check the /data/data/YOUR_APP_NAME/cache/ folder for the rn_image_picker_lib_temp files and download them to your computer.

Use an exif viewer and notice that the exif data is empty.

- step to reproduce the issue
As Above

### Solution
- Add useLegacyPicker flag which invokes INTENT.ACTION_PICK when enabled.
- Copy EXIF data when resizing image.
 
What needs to be done to address this issue?
See code fix.

### Additional Information

- Image Picker version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
- React Native version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
- Platform: [FILL THIS OUT: iOS, Android, or both?]
- Development Operating System: [FILL THIS OUT: Are you developing on MacOS, Linux, or Windows?]
- Dev tools: [FILL THIS OUT: Xcode or Android Studio version, iOS or Android SDK version, if applicable] navigation crash for Android

PR #2365, opened 2025-04-07
Add UseLegacyPicker option and allow resize of image to retain exif data

DylanJCraggs:
https://github.com/react-native-image-picker/react-native-image-picker/issues/2364


jtklein:
Compare: https://github.com/react-native-image-picker/react-native-image-picker/pull/2237

PR #2401, opened 2026-02-26
Fix duplicate video files when multiple videos share the same filename

petercvek:
## Summary

When selecting multiple videos from the iOS photo library that share the same filename (e.g. `FullSizeRender.mov`), `mapVideoToAsset` copies each video to the same temp directory path, causing each file to overwrite the previous one. All returned asset URIs then point to the same physical file (the last one copied).

This commonly happens when users edit/trim multiple videos in the iOS Photos app, as iOS names all edited video exports `FullSizeRender.mov`.

### Before
```
Video 1 → /tmp/FullSizeRender.mov  (overwritten by Video 2)
Video 2 → /tmp/FullSizeRender.mov  (overwritten by Video 3)
Video 3 → /tmp/FullSizeRender.mov  ← only this survives
```
All 3 URIs point to Video 3.

### After
```
Video 1 → /tmp/FullSizeRender_A1B2C3D4-...-E5F6.mov
Video 2 → /tmp/FullSizeRender_G7H8I9J0-...-K1L2.mov
Video 3 → /tmp/FullSizeRender_M3N4O5P6-...-Q7R8.mov
```
Each URI points to the correct file.

## Changes

Append a UUID to the temp filename in `mapVideoToAsset` to ensure uniqueness. The original filename is still returned in the response's `fileName` field for display purposes.

This is consistent with:
- The `formatAsMp4` branch in the same method, which already uses `[[NSUUID UUID] UUIDString]` for the output path
- The Android implementation (`Utils.createFile`), which uses `UUID.randomUUID()` for temp filenames


gevorgsharpant (approved):

raid5:
Any update on this?

